PR #2, opened 2017-09-20
[46] Residents can locate problem

leeky:
A resident can visit `/address_search/` and enter their postcode. If it is in the correct format, they are shown a list of matching addresses.

If they have JavaScript enabled, the submit buttons are hidden until they select an address from the list. 

This is a first attempt at this, and will need revisiting in the future (clicking back button, shows radio button previously selected, but the submit button is still hidden, for example). I'm suspecting this should be some standalone JS tests (perhaps using `jasmine-rails`), as this behaviour should be decoupled from the app.

NB: Currently this entirely returns canned responses in place of the API, which is still being built.

### DIFF
--- a/app/assets/javascripts/form.js
+++ b/app/assets/javascripts/form.js
@@ -1,0 +1,19 @@
+$(document).ready(function() {
+  $('body').addClass('js-enabled');
+
+  // Loop over each form on the page
+  $('form').each(function() {
+
+    // If the form contains radio buttons, hide the submit button
+    if ($(this).find('input[type=radio]').length > 0) {
+      $(this).find('input[type=button]').hide();
+      $(this).find('input[type=submit]').hide();
+    }
+
+    // Clicking on any radio button will show the submit button
+    if ($(this).find('input[type=radio]').click(function() {
+      $(this).closest('form').find('input[type=button]').show();
+      $(this).closest('form').find('input[type=submit]').show();
+    }));
+  });
+});
--- a/app/controllers/address_searches_controller.rb
+++ b/app/controllers/address_searches_controller.rb
@@ -1,11 +1,15 @@
 class AddressSearchesController < ApplicationController
   def show
+    @address_search = AddressSearch.new
+  end
+
+  def update
     @address_search = AddressSearch.new(address_search_params[:address_search])
 
-    first_visit = address_search_params.empty?
-    @show_results = !first_visit && @address_search.valid?
-
-    return unless @show_results
+    unless @address_search.valid?
+      render :show
+      return
+    end
 
     address_finder = AddressFinder.new(HackneyApi.new)
     @address_search_results = address_finder.find(@address_search)
--- a/app/controllers/address_searches_controller.rb
+++ b/app/controllers/address_searches_controller.rb
@@ -1,8 +1,19 @@
 class AddressSearchesController < ApplicationController
   def show
-    @address_search = AddressSearch.new
-    @address = Address.new
+    @address_search = AddressSearch.new(address_search_params[:address_search])
+
+    @first_visit = address_search_params.empty?
+    return if @first_visit
+
     address_finder = AddressFinder.new(HackneyApi.new)
     @address_search_results = address_finder.find(@address_search)
+
+    @address = Address.new
+  end
+
+  private
+
+  def address_search_params
+    params.permit(address_search: [:postcode])
   end
 end
--- a/app/controllers/address_searches_controller.rb
+++ b/app/controllers/address_searches_controller.rb
@@ -2,5 +2,6 @@ class AddressSearchesController < ApplicationController
   def show
     @address_search = AddressSearch.new
     @address = Address.new
+    @address_search_results = AddressFinder.new.find(@address_search)
   end
 end
--- a/app/controllers/address_searches_controller.rb
+++ b/app/controllers/address_searches_controller.rb
@@ -1,13 +1,13 @@
 class AddressSearchesController < ApplicationController
-  def show
+  def new
     @address_search = AddressSearch.new
   end
 
-  def update
+  def create
     @address_search = AddressSearch.new(address_search_params[:address_search])
 
     unless @address_search.valid?
-      render :show
+      render :new
       return
     end
 
--- a/app/controllers/address_searches_controller.rb
+++ b/app/controllers/address_searches_controller.rb
@@ -2,6 +2,7 @@ class AddressSearchesController < ApplicationController
   def show
     @address_search = AddressSearch.new
     @address = Address.new
-    @address_search_results = AddressFinder.new.find(@address_search)
+    address_finder = AddressFinder.new(HackneyApi.new)
+    @address_search_results = address_finder.find(@address_search)
   end
 end
--- a/app/controllers/address_searches_controller.rb
+++ b/app/controllers/address_searches_controller.rb
@@ -1,0 +1,6 @@
+class AddressSearchesController < ApplicationController
+  def show
+    @address_search = AddressSearch.new
+    @address = Address.new
+  end
+end
--- a/app/controllers/address_searches_controller.rb
+++ b/app/controllers/address_searches_controller.rb
@@ -2,8 +2,10 @@ class AddressSearchesController < ApplicationController
   def show
     @address_search = AddressSearch.new(address_search_params[:address_search])
 
-    @first_visit = address_search_params.empty?
-    return if @first_visit
+    first_visit = address_search_params.empty?
+    @show_results = !first_visit && @address_search.valid?
+
+    return unless @show_results
 
     address_finder = AddressFinder.new(HackneyApi.new)
     @address_search_results = address_finder.find(@address_search)
--- a/app/controllers/addresses_controller.rb
+++ b/app/controllers/addresses_controller.rb
@@ -1,0 +1,5 @@
+class AddressesController < ApplicationController
+  def update
+    redirect_to new_appointment_path
+  end
+end
--- a/app/controllers/addresses_controller.rb
+++ b/app/controllers/addresses_controller.rb
@@ -1,8 +1,8 @@
 class AddressesController < ApplicationController
-  def update
+  def create
     if property_reference.empty?
-      return redirect_back(
-        fallback_location: address_search_path,
+      return redirect_to(
+        new_address_search_path,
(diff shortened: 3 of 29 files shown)


spec/features/resident_can_locate_problem_spec.rb
```diff
@@ -60,7 +60,16 @@
...
+    fill_in :address_search_postcode, with: 'NNN1 1AAA'
+    click_button t('helpers.submit.address_search.create')
+
+    expect(page).to have_content "Postcode doesn't seem to be valid"
+  end
+
+  scenario 'when no postcode is entered' do
+    visit '/address_search/'
+
+    fill_in :address_search_postcode, with: '  '
```
dgmstuart: I'd normally only test _one_ validation failure in a feature spec, the idea being that that is sufficient to test that everything integrates correctly. 

No harm in more - the tradeoff is with speed.

spec/models/address_search_spec.rb
```diff
@@ -24,6 +24,10 @@
       params = { postcode: 'n16 8qr' }
       expect(AddressSearch.new(params).data).to eq(postcode: 'N16 8QR')
     end
+
+    it 'returns the search data even if no params were provided' do
```
dgmstuart: I can't work out why this is - I've checked and nothing seems to break if it's not true.

Is my understanding correct?: if no params are provided then the form is invalid, and if the form is invalid then the caller shouldn't be asking it for data.

If that's the case then would it be better to raise an exception in this scenario?

app/controllers/address_searches_controller.rb
```diff
@@ -1,11 +1,15 @@
 class AddressSearchesController < ApplicationController
   def show
-    @address_search = AddressSearch.new(address_search_params[:address_search])
```
dgmstuart: Is this still "showing" an address search, or creating a new one?

Would it be better for these two actions to be the normal pairs? new/create (or edit/update?)

app/controllers/address_searches_controller.rb
```diff
@@ -1,11 +1,15 @@
...
-    first_visit = address_search_params.empty?
-    @show_results = !first_visit && @address_search.valid?
+  def update
+    @address_search = AddressSearch.new(address_search_params[:address_search])
 
-    return unless @show_results
+    unless @address_search.valid?
+      render :show
+      return
+    end
```
dgmstuart: How would it look if the happy case redirected to an "index" action on the Addresses controller instead?

app/views/address_searches/update.html.haml
```diff
@@ -0,0 +1,16 @@
+- if @address_search.errors
+  = @address_search.errors.full_messages.join(', ')
+
+= simple_form_for @address_search, url: address_search_path, method: :get do |f|
+  = @address_search.postcode
+  = f.button :submit, 'Change'
```
dgmstuart: Since this doesn't have any fields, could it be a link instead of a form?

app/controllers/addresses_controller.rb
```diff
@@ -1,13 +1,16 @@
 class AddressesController < ApplicationController
   def update
-    selected_answers_store = SelectedAnswerStore.new(session)
-
-    property_reference = update_params[:property_reference]
+    if property_reference.empty?
+      return redirect_back(
+        fallback_location: address_search_path,
+        alert: 'Please choose your address to continue'
```
dgmstuart: Should this alert string go in the locales?

app/controllers/addresses_controller.rb
```diff
@@ -1,13 +1,16 @@
 class AddressesController < ApplicationController
   def update
-    selected_answers_store = SelectedAnswerStore.new(session)
```
dgmstuart: This commit says "TODO" - does that mean you intended to squash it in with another commit?

app/controllers/addresses_controller.rb
```diff
@@ -17,4 +20,8 @@ def update
   def update_params
     params.require(:address).permit(:property_reference)
   end
+
+  def property_reference
+    update_params[:property_reference]
+  end
```
dgmstuart: Is there a benefit to extracting this function? The tradeoff is that it hides away useful information at the bottom of the file and as a controller grows it can become unclear which actions the function is used by.

I know it's a common pattern from TVH, but I'm not sure it's a good one!

spec/features/resident_can_locate_problem_spec.rb
```diff
@@ -90,4 +90,24 @@
...
+
+  scenario 'performing a search, but not selecting any address' do
+    matching_property = {
+      'property_reference' => 'abc123',
+      'short_address' => 'Flat 1, 8 Hoxton Square, N1 6NU',
+    }
+
+    fake_api = instance_double(JsonApi)
+    allow(fake_api).to receive(:get).with('properties?postcode=N1 6NU').and_return([matching_property])
+    allow(fake_api).to receive(:get).with('properties/').and_return({})
```
dgmstuart: Is this line required? The test seems to pass without it.

app/controllers/addresses_controller.rb
```diff
@@ -1,13 +1,16 @@
 class AddressesController < ApplicationController
   def update
-    selected_answers_store = SelectedAnswerStore.new(session)
-
-    property_reference = update_params[:property_reference]
+    if property_reference.empty?
```
dgmstuart: Would this make sense as a validation on a form object - or on the Address model?

app/assets/javascripts/form.js
```diff
@@ -0,0 +1,18 @@
+$(document).ready(function() {
+  $('body').addClass('js-enabled');
+
+  // Loop over each form on the page
+  $('form').each(function() {
+
+    // If the form contains radio buttons, hide the submit button
+    if ($(this).find('input[type=radio]').length > 0) {
+      $(this).find('input[type=submit]').hide();
```
dgmstuart: Do we also need to account for `input[type=button]`, or will we just avoid using that?

spec/features/resident_can_locate_problem_spec.rb
```diff
@@ -91,7 +91,32 @@
...
-  scenario 'performing a search, but not selecting any address' do
+  scenario 'performing a search and not selecting an address, prevents form submission', js: true do
+    matching_property = {
+      'property_reference' => 'abc123',
+      'short_address' => 'Flat 1, 8 Hoxton Square, N1 6NU',
+    }
+
+    fake_api = instance_double(JsonApi)
+    allow(fake_api).to receive(:get).with('properties?postcode=N1 6NU').and_return([matching_property])
+    allow(fake_api).to receive(:get).with('properties/').and_return({})
```
dgmstuart: As per previous comment - I think this line might not be required?

spec/features/resident_can_locate_problem_spec.rb
```diff
@@ -91,7 +91,32 @@
...
+    fill_in :address_search_postcode, with: 'N1 6NU'
+    click_button t('helpers.submit.address_search.create')
+
+    expect(page).to have_no_button t('helpers.submit.address.create')
+
+    within '#address-search-results' do
+      choose 'Flat 1, 8 Hoxton Square, N1 6NU'
+    end
+
+    expect(page).to have_button t('helpers.submit.address.create')
```
dgmstuart, line 115: Nice test 👍

dgmstuart (changes requested):
The approach looks great and it seems to work like a charm. I have a bunch of questions about the code.

Not sure why all my comments showed up as being on old commits - maybe some rebasing went on?

Also I'd like to squash in the FIX commits I made previously - or reword them if you think they belong as first class citizens.

spec/features/resident_can_locate_problem_spec.rb
leeky: Good spot. I think this was left over from some earlier testing :+1:

app/controllers/addresses_controller.rb
leeky: I agree with you. The change was only necessary to keep RuboCop happy. 

We should consider allowing more lines per methods, perhaps?

app/controllers/addresses_controller.rb
leeky: I've removed it and added an additional card in the backlog.

dgmstuart:
@leeky 

As discussed, the following spec will fail:
```ruby 
it 'adds a space in the middle' do
  params = { postcode: 'N168QR' }
  expect(AddressSearch.new(params).data).to eq(postcode: 'N16 8QR')
end
```

I'm now feeling like we should have two separate objects: 

- a Form object which is responsible for validation 
- a Formatter object which will need to be used (by coincidence) to present data to both the view and the query. This should expect a valid postcode and should raise some error if the data it's given can't be formatted as a postcode.

dgmstuart:
rebasing

spec/features/resident_can_locate_problem_spec.rb
dgmstuart: Done

spec/models/address_search_spec.rb
dgmstuart: Was removed

app/controllers/address_searches_controller.rb
dgmstuart: (changed)

app/controllers/address_searches_controller.rb
dgmstuart: (discussed - that would mean having to handle the case where that page is viewed without a postcode having been selected)

app/views/address_searches/update.html.haml
dgmstuart: (changed)

app/controllers/addresses_controller.rb
dgmstuart: (maybe, but this works fine)
This should be `blank?` though

app/views/address_searches/update.html.haml
dgmstuart: Done

dgmstuart (approved):

app/models/address_search.rb
```diff
@@ -12,7 +12,7 @@ def data
   private
 
   def normalised_postcode
-    postcode.strip.gsub(/\s+/, ' ').upcase
+    postcode.strip.gsub(/\s+/, '').insert(-4, ' ').upcase
```
leeky: Didn't know about `insert` :+1:

PR #3, opened 2017-09-25
(63) Start page telling users about the service

leeky:
Take the start page from the prototype and make it the root page of the service.

 - I've renamed the temporary `WelcomeController` to `StartController`.
 - This also adds basic styling, which is mainly GOVUK styling with some of the styling from Hackney, most notably the header and footer. NB: This is a first pass - the UI will get more love as I start building the forms.

start-page


dgmstuart:
Looks great 😄 👍

dgmstuart (approved):

PR #5, opened 2017-09-28
(65) Users can answer a repair question

leeky:
This builds the first question in the decision tree, and is heavily influenced by [GOVUK Verify](https://github.com/alphagov/verify-frontend).

Each question comprises:

 - Controller in `app/controllers/questions/`
 - Form object in `app/models/`, which handles validation
 - Form view in `app/models/questions/`, to display the form to the user
 - I18n entries in `config/locales/` which are used by `simple_form` to populate the form options
 - A pair of Non-RESTful routes in `config/routes.rb`.

Additionally, I've added the `i18n-debug` gem, which logs any I18n key lookups, to aid development because Rails' choice of i18n keys is often non-obvious and hard to debug!


app/controllers/questions/start_controller.rb
```diff
@@ -22,5 +24,11 @@ def submit
     def start_form_params
       params.require(:start_form).permit!
     end
+
+    def store_form_in_session!
+      SelectedAnswerStore
+        .new(session)
+        .store_selected_answers(:priority_repair, @form.priority_repair)
```
dgmstuart: Would it be more explicit/safer to pass in the form as a parameter here? 

If we were in a simple class and `@form` was set in an instance variable then it would always definitely have a value, but in a controller it won't always have a value.

Alternatively, how would you feel about increasing the line length in rubocop and inlining it?

spec/controllers/questions/start_controller_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe Questions::StartController do
+  describe '#submit' do
+    it 'stores valid responses to the session' do
+      post :submit, params: { start_form: { priority_repair: 'Yes' } }
```
dgmstuart: Is this the value it actually gets from the view or will it be lowercase?

app/controllers/questions/start_controller.rb
```diff
@@ -22,5 +24,11 @@ def submit
     def start_form_params
       params.require(:start_form).permit!
     end
+
+    def store_form_in_session!
```
dgmstuart: Since this is a question of "continue or exit", do we need to store this value in the session? How will we use it later?

app/models/start_form.rb
```diff
@@ -1,10 +1,14 @@
 class StartForm
   include ActiveModel::Model
 
-  attr_accessor :priority_repair
+  attr_reader :priority_repair
```
dgmstuart: Love it 👍

dgmstuart (changes requested):
Looks good - some comments mostly on one commit.

spec/controllers/questions/start_controller_spec.rb
leeky: Good spot :+1:

app/controllers/questions/start_controller.rb
leeky: Storing the session got bumped out of this card, but I'd already implemented it here, even though it doesn't make sense in this context.

How about I remove saving into the session here altogether, and revisit it when there's a need to do so?

app/controllers/questions/start_controller.rb
leeky: See below.

leeky:
@dgmstuart As discussed, I've removed the session saving code and associated test, so this should be ready for another look.

dgmstuart (approved):

dgmstuart:
LGTM 👍

dgmstuart:
rebased

PR #6, opened 2017-09-28
(98) Style address selection

leeky:
The radio buttons for address selection are now rendered in the correct styling.

**NB: This branch is currently based off #5, as that branch includes the relevant CSS radio button styles from GOVUK elements.**

## Before
before

## After
after

### DIFF
--- a/spec/support/capybara_helpers.rb
+++ b/spec/support/capybara_helpers.rb
@@ -1,0 +1,5 @@
+module CapybaraHelpers
+  def choose_radio_button(label)
+    find(:label, text: label).click
+  end
+end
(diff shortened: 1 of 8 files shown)


dgmstuart:
rebased

spec/support/capybara_helpers.rb
```diff
@@ -0,0 +1,5 @@
+module CapybaraHelpers
+  def choose_radio_button(label)
```
dgmstuart, line 2: Nice 👍

dgmstuart (changes requested):
As discussed, let's include the "Address Search" page styling in this task, and both these pages need a title heading

leeky:
@dgmstuart As discussed, I've now styled the two forms so they now have the correct titles, buttons etc.

I've also rebased against develop.

dgmstuart:
Looks great 👍

dgmstuart (approved):

PR #7, opened 2017-09-28
[77] Connect to the Hackney API

dgmstuart:
Not much to say: we went around the houses on this one, but in the end, stubbing out the Json API feels right

Decided not to go down the route of a .env.example or an intializer which checks that it's set. Would be happy to take either approach in future though.


spec/spec_helper.rb
```diff
@@ -58,7 +58,7 @@
   #   # Allows RSpec to persist some state between runs in order to support
   #   # the `--only-failures` and `--next-failure` CLI options. We recommend
   #   # you configure your source control system to ignore this file.
-  #   config.example_status_persistence_file_path = "spec/examples.txt"
+  config.example_status_persistence_file_path = 'spec/failures.txt'
```
leeky, line 62: Never know why this isn't on by default :+1:

spec/features/resident_can_locate_problem_spec.rb
```diff
@@ -76,6 +76,11 @@
   end
 
   scenario 'changing the postcode after searching' do
+    fake_api = instance_double(JsonApi)
```
leeky, line 79: If we don't care about the response from the API, could these three lines be replaced with the following..?

```
fake_api = instance_double(JsonApi, get: [])
```

spec/queries/json_api_spec.rb
```diff
@@ -1,33 +1,58 @@
...
 
         result = json_api.get('properties/zxc098')
 
-        expect(result).to be_a(Hash)
-        expect(result.keys).to include('property_reference', 'short_address', 'uprn')
+        expect(result).to eq('uprn' => 'xyz987', 'property_reference' => 'abc123', 'short_address' => '1 some road')
+      end
+    end
+
+    context 'when the api root is missing' do
```
leeky, line 70: Should there also be a test that ensures that `api_root` is a valid URL?

Currently a typo such as `htp://hackney.api:8080/` wouldn't raise an error until a request is made against the connection.

app/queries/json_api.rb
```diff
@@ -1,16 +1,20 @@
...
       conn.adapter Faraday.default_adapter
-      conn.response :json, content_type: /\bjson$/
+      conn.response :json
     end
   end
 
   def get(path)
     @connection.get(path).body
+  rescue Faraday::ParsingError => e
+    raise InvalidResponseError, e.message
```
leeky: :+1:

leeky (commented):
Just a few comments, but otherwise looks good!

spec/features/resident_can_locate_problem_spec.rb
dgmstuart: Could do, but with my Ben hat on, I feel like there's value in the 'symmetry' between the test cases: the philosophy is that if they're all built in the same way then it's easier to see what the differences are.

spec/queries/json_api_spec.rb
dgmstuart: Good shout - will do

spec/features/resident_can_locate_problem_spec.rb
leeky: Good point 👍

spec/queries/json_api_spec.rb
dgmstuart: As discussed - I don't think there's a way to find errors in the URI in general, since URI.parse will accept most strings.

leeky (approved):
LGTM! :shipit:

PR #8, opened 2017-09-29
(45) Unidentified addresses directed to call centre

leeky:
When the resident cannot see their address in the address lookup search results, give them the option to choose "My address isn't here". When the form is submitted, they are sent to a page that asks them to call Hackney.

cannot-find-address

**NB: This relies on #6, which should be merged first**

### DIFF
--- a/app/controllers/addresses_controller.rb
+++ b/app/controllers/addresses_controller.rb
@@ -7,6 +7,10 @@ class AddressesController < ApplicationController
       )
     end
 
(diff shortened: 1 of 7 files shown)


dgmstuart:
Rebased

dgmstuart (approved):
LGTM - happy for this to be merged when the dependent PRs are resolved

leeky:
Rebased and fixed conflicts.

PR #10, opened 2017-10-02
(79) Problem description page (known)

leeky:
Create a dummy form for entering details about the repair.

NB: Anything saved into the form is currently **NOT** persisted. Additionally, submitting the form redirects the user to the address search page.

As part of this branch, I've included the GOVUK focus styles (which highlight the currently-focussed form element) and broken the stylesheet up into smaller parts.

From: `/questions/describe-repair` 

describe-repair

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,5 +16,8 @@ Rails.application.routes.draw do
       to: 'pages#address_isnt_here',
       as: 'address_isnt_here'
 
+  get '/describe-repair', to: 'describe_repair#index', as: 'describe_repair'
+  post '/describe-repair', to: 'describe_repair#submit'
+
   root to: 'start#index'
 end
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -11,39 +11,6 @@
  * It is generally better to create a new file per style scope.
  *
  *= require_self
- *= require_tree .
+ *= require hackney
  */
 
-@import 'variables';
-
-/* From GDS's alphagov/govuk_frontend_toolkit */
-@import 'font_stack';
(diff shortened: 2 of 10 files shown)


config/routes.rb
```diff
@@ -6,6 +6,9 @@
   namespace :questions do
     get '/start', to: 'start#index', as: 'start'
     post '/start', to: 'start#submit', as: 'start_submit'
+
+    get '/describe-repair', to: 'describe_repair#index', as: 'describe_repair'
+    post '/describe-repair', to: 'describe_repair#submit'
```
dgmstuart: Does this belong under the 'questions' path, or is it its own thing? Maybe it doesn't matter at this point - easy to change later?

app/controllers/questions/describe_repair_controller.rb
```diff
@@ -0,0 +1,10 @@
+module Questions
+  class DescribeRepairController < ApplicationController
+    def index; end
+
+    def submit
+      # redirect_to 'next_form'
```
dgmstuart: Should this redirect to the address search page?

dgmstuart (changes requested):
Looks good - I think we can join this up to the next page?

Stylesheet cleanup is nice 👍

config/routes.rb
leeky: Possibly. As this is a named route it'd be easy to change in the future.

leeky: On second thoughts, I agree with you. I'll move these out of the 'questions' namespace.

dgmstuart (approved):

dgmstuart:
LGTM 👍

PR #11, opened 2017-10-02
[104] allow longer methods in controllers

dgmstuart:
We in general want our methods to be short, but this is hard to do in
Controllers without prematurely extracting objects and/or mangling the
code.

Rubocop doesn't support specifying different rules for different
directories in the main .rubocop.yml, but additional rubocop.yml files
in directories apply to all the files in and below that directory


leeky (approved):
LGTM! :shipit:

PR #12, opened 2017-10-02
(80) Problem description page (unknown)

leeky:
Add an additional dummy form, similar to the one in #10. This one is for the "problem is something else" path, so the wording is slightly different.

I'd also noticed that the `questions/start` page had slightly different markup to the other forms I've built so far, so I've fixed that here.

From: `/questions/describe-unknown-repair`:

describe-unknown-repair

### DIFF
--- a/app/controllers/describe_unknown_repair_controller.rb
+++ b/app/controllers/describe_unknown_repair_controller.rb
@@ -1,0 +1,7 @@
+class DescribeUnknownRepairController < ApplicationController
+  def index; end
+
+  def submit
+    redirect_to new_address_search_path
+  end
+end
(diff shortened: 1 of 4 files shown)


app/controllers/questions/describe_unknown_repair_controller.rb
```diff
@@ -0,0 +1,10 @@
+module Questions
+  class DescribeUnknownRepairController < ApplicationController
+    def index; end
+
+    def submit
+      # redirect_to 'next_form'
```
dgmstuart: Should this also redirect to the Address Search page?

dgmstuart (changes requested):
Looks good, but I think we can join this page up into the flow already

app/controllers/questions/describe_unknown_repair_controller.rb
leeky: Yep. Updated.

dgmstuart (approved):

PR #14, opened 2017-10-02
[82] Confirmation page (dummy)

dgmstuart:
PLEASE CHECK the copy for sanity and that we have the right set of data on this page.

This is mostly hardcoded - except for the repair reference.

Mocked up a Confirmation object as a Struct because I don't want to be
passing 20 individual instance variables to the view ;)

The spec here was just to drive out the routing and controller. When we
have real templates for everything, I anticipate this being a full
end-to-end test which goes through all the screens and checks that the
correct values are shown on the confirmations page.

Need to remember to pull out the phone number to i18n at some point

### DIFF
--- a/app/assets/stylesheets/components/main.scss
+++ b/app/assets/stylesheets/components/main.scss
@@ -19,11 +19,25 @@ main {
 
   h2 {
     @include heading-24;
+
+    font-weight: bold;
+  }
+
+  h3 {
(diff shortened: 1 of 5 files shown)


leeky (approved):
As discussed, I've made some styling tweaks to make the page look the same as the others on the site.

Otherwise, LGTM! :shipit:

PR #15, opened 2017-10-03
(103) Fix the responsiveness of the site

leeky:
A few tweaks to improve the experience on mobile devices.

 - Include `viewpoint` meta tag which managed page scaling on mobile
 - `http-equiv: x-ua-compatible` ensures that Internet Explorer usees the most up-to-date renderer.
 - Use high-DPI logo that looks nicer on phone/tablet/retina displays.
 - Adjust line height of footer elements when they wrap across multiple lines.

Now that we're including the GOVUK styles properly, the pages automatically switch to the smaller font size on mobile, and all content is surrounded by whitespace so it isn't pushed to the sides of the screen.


dgmstuart (approved):
Still seems to have issues at iPhone5 size (according to Chrome) but it's definitely an improvement and we have bigger fish to fry right now

PR #16, opened 2017-10-03
(86) 'Back' links throughout the form

dgmstuart:
**Merge https://github.com/LBHackney-IT/maintain-my-home/pull/17 first**

The first step of this was to glue all the pages together so that we can now click through the form (nearly) end-to-end and back again.

There's a lot of duplication in the specs - it would be nice to get rid of that somehow, but although it's currently possible to jump to arbitrary points in the form (which would be one way to shortcut this), that won't be true in future (when we check what questions you've already answered), so I don't want to cause us future pain by creating specs which will break later.

This is currently hardcoded per controller. We'll need to do something special for branching and rejoining flows, but currently there aren't any through the app (e.g. there's no way to get to the unknown repair description page) so I propose we deal with that later and implement the pending specs which I've left at that point.

In any case - there's already a LOT here :scream:

### DIFF
--- a/app/presenters/back.rb
+++ b/app/presenters/back.rb
@@ -12,11 +12,7 @@ class Back
   private
 
   def text
-    "#{prefix} #{page_description}"
-  end
-
-  def prefix
-    I18n.t 'back_links.prefix'
+    I18n.t 'back_links.prefix', page_description: page_description
   end
 
   def page_description
--- a/app/presenters/back.rb
+++ b/app/presenters/back.rb
@@ -1,0 +1,25 @@
+class Back
+  def initialize(controller_name:, routes: Rails.application.routes)
+    @controller_name = controller_name
+    @helpers = ActionController::Base.helpers
+    @routes = routes
+  end
+
+  def link
+    @helpers.link_to text, @routes.path_for(controller: @controller_name)
+  end
+
+  private
+
+  def text
+    "#{prefix} #{page_description}"
+  end
+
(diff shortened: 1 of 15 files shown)


leeky:
Rebasing

app/presenters/back.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def link
+    @helpers.link_to text, @routes.path_for(controller: @controller_name)
+  end
+
+  private
+
+  def text
+    "#{prefix} #{page_description}"
```
leeky: You could also generate this text using I18n interpolation, e.g in `en.yml`:

`Back to %{page_description}`

Which is invaluable if we were translating to a language where "Back to" should be last in the sentence.

dgmstuart: Good shout. I was wondering about that. I guess it makes no difference that the parameter passed to it will be another I18n string.

dgmstuart: @leeky pushed a commit for this

leeky (approved):
LGTM!

@dgmstuart Could you rebase, and then I'm happy for this to be merged.

dgmstuart:
A sadly painful rebase. Ready for another look. One new commit (2c961cb) and amended 7cd9e8b to add in the `@back` object in the case where the AddressForm is invalid. 

On the plus side, the existing tests told us something was wrong here, which gives me even more confidence in our test suite.

Ready for another look

PR #17, opened 2017-10-04
(110) Link all the pages together

dgmstuart:
This is an intermediate step: we built all the pages, but haven't actually linked them together yet. Required a bit of re-jigging.

I've taken the approach that where bits of the flow don't exist yet, it's better to be able to go all the way through for now, and then we can insert new pages later.


config/routes.rb
```diff
@@ -1,5 +1,7 @@
 Rails.application.routes.draw do
-  resource :address_search, only: %i[new create]
+  get '/address-search', to: 'address_searches#index', as: 'address_search'
```
leeky, line 2: :+1:

leeky (approved):
LGTM! :shipit:

PR #18, opened 2017-10-04
(108) Standardise Address Selection submit button

dgmstuart:
- The other stages of the form all say 'Continue'
- 'Use this address' doesn't make sense for the "My address isn't here"
  option

Hooray for locales!

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -11,13 +11,10 @@ en:
     no_selection: Please choose an answer before continuing.
   helpers:
     submit:
-      address:
-        create: Use this address
       address_search:
         create: Find my address
       create: Continue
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -12,7 +12,7 @@ en:
   helpers:
     submit:
       address:
-        create: Use this address
+        create: Continue
```
leeky: The default button (under `simple_form.helpers.submit.create`) is already `Continue`, so you can remove this key altogether.

(*Double* hooray locales!)

dgmstuart:
Amended the commit - ready for another look

leeky (approved):
LGTM!

PR #19, opened 2017-10-04
(91) Show optional fields and validation errors

leeky:
- Forms now indicate when fields are optional rather than when they are required
 - The 'describe unknown repair' and 'contact details' forms have required fields, so now have form objects to perform validation
 - Validation prevents incomplete forms being submitted
 - Validation errors are in the GOVUK style
 -  Optional fields need to have labels

validation-errors

NB: GOVUK style suggested having a series of links to the invalid fields at the top of the page. This is quite hard to do because it requires some additional ARIA attributes on the form and error fields. I've therefore omitted this for now, which feels fine since we're avoiding too many questions per page.

### DIFF
--- a/app/controllers/addresses_controller.rb
+++ b/app/controllers/addresses_controller.rb
@@ -1,18 +1,20 @@
 class AddressesController < ApplicationController
   def create
-    if property_reference.blank?
-      return redirect_to(
-        address_search_path,
-        alert: t('addresses.errors.blank')
-      )
+    @form = AddressForm.new(address_form_params)
+
+    if @form.invalid?
+      @address_search = AddressSearch.new(postcode: @form.postcode)
+
+      address_finder = AddressFinder.new(HackneyApi.new)
+      @address_search_results = address_finder.find(@address_search)
+
+      return render 'address_searches/create'
     end
 
-    if property_reference == 'address_isnt_here'
-      return redirect_to address_isnt_here_path
-    end
+    return redirect_to address_isnt_here_path if @form.address_isnt_here?
 
     api = HackneyApi.new
-    address = api.get_property(property_reference: property_reference)
+    address = api.get_property(property_reference: @form.property_reference)
 
     SelectedAnswerStore.new(session).store_selected_answers('address', address)
 
@@ -21,11 +23,7 @@ class AddressesController < ApplicationController
 
   private
 
-  def update_params
-    params.require(:address).permit(:property_reference)
-  end
-
-  def property_reference
-    update_params[:property_reference]
+  def address_form_params
+    params.require(:address_form).permit(:property_reference, :postcode)
   end
 end
--- a/app/controllers/contact_details_controller.rb
+++ b/app/controllers/contact_details_controller.rb
@@ -2,9 +2,23 @@ class ContactDetailsController < ApplicationController
   def index
     selected_answer_store = SelectedAnswerStore.new(session)
     @selected_answers = selected_answer_store.selected_answers['address']
+
+    @form = ContactDetailsForm.new
   end
 
   def submit
+    selected_answer_store = SelectedAnswerStore.new(session)
+    @selected_answers = selected_answer_store.selected_answers['address']
+
+    @form = ContactDetailsForm.new(contact_details_form_params)
+    return render :index unless @form.valid?
+
     redirect_to confirmation_path('abc123')
+  end
+
+  private
+
+  def contact_details_form_params
+    params.require(:contact_details_form).permit!
   end
 end
--- a/config/initializers/simple_form.rb
+++ b/config/initializers/simple_form.rb
@@ -1,3 +1,8 @@
+require Rails.root.join('lib', 'simple_form', 'optional_labels')
+
+SimpleForm::Components::Labels.prepend SimpleForm::Components::OptionalLabels
+SimpleForm::Components::Labels::ClassMethods.include SimpleForm::Components::OptionalLabels::ClassMethods
+
 SimpleForm.setup do |config|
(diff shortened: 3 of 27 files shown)


spec/models/contact_details_form_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe ContactDetailsForm do
+  describe 'validations' do
+    it '.full_name' do
```
dgmstuart: I guess 'it validates presence of full name' is more descriptive but more verbose?
This is fine for now, but at some point we're going to need additional validations - e.g. field length

spec/models/contact_details_form_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe ContactDetailsForm do
+  describe 'validations' do
```
dgmstuart, line 4: Do we need a spec that shows that this is valid when all fields are present? I think currently this would pass if `valid?` was always false.

spec/models/contact_details_form_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    it '.telephone_number' do
+      form = ContactDetailsForm.new
+      form.valid?
+
+      expect(form.errors.details[:telephone_number]).to include(error: :blank)
+    end
+
+    it '.callback_time' do
```
dgmstuart: Would there be value in demonstrating that this form is valid when only one checkbox is checked?

config/initializers/simple_form.rb
```diff
@@ -47,4 +52,7 @@
 
   # Define the default class of the input wrapper of the boolean input.
   config.boolean_label_class = 'checkbox'
+
+  # Append 'optional' label instead of prepending
+  config.label_text = ->(label, required, _explicit_label) { "#{label} #{required}" }
```
dgmstuart: Are we able to rename this parameter to `optional`? Would that make any sense?

dgmstuart: Also: this will result in a space after the label for required fields. Probably no big deal?
It made sense in the blog post because they were marking both required and optional fields.

spec/features/users_cannot_submit_incomplete_forms_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    within '.contact_details_form_telephone_number' do
+      expect(page).to have_css('span.error-message')
+      expect(page).to have_css('div input[aria-invalid=true]')
+    end
+
+    within '.contact_details_form_callback_time' do
+      expect(page).to have_css('span.error-message')
+      expect(page).to have_css('div input[aria-invalid=true]')
+    end
```
dgmstuart, line 51: Do we now need further steps in this spec to show that after these validation errors we can go on to successfully submit a form?

app/controllers/contact_details_controller.rb
```diff
@@ -2,9 +2,23 @@ class ContactDetailsController < ApplicationController
   def index
     selected_answer_store = SelectedAnswerStore.new(session)
     @selected_answers = selected_answer_store.selected_answers['address']
+
+    @form = ContactDetailsForm.new
   end
 
   def submit
+    selected_answer_store = SelectedAnswerStore.new(session)
+    @selected_answers = selected_answer_store.selected_answers['address']
```
dgmstuart, line 11: Do we need @selected_answers in the happy path case?
I'm never quite sure if it's worth making the distinction in this sort of situation.

dgmstuart:
The validations on the first page don't seem to be styled (yes, I had to force the submit button to be displayed using the inspector):
screen shot 2017-10-04 at 16 52 14

dgmstuart:
Ditto the Address selection page, which also jumps back to the address search page instead of re-rendering the address selection list:

screen shot 2017-10-04 at 16 54 37

dgmstuart (changes requested):
Generally looks great - the errors on individual fields look really nice.

The main thing missing is specs which test the forms being submitted successfully.

app/controllers/contact_details_controller.rb
leeky: I did this so that the `index` and `submit` methods were similar shaped, rather than having the `SelectedAnswerStore` being in the conditional.

spec/models/contact_details_form_spec.rb
leeky: :+1: Added

config/initializers/simple_form.rb
leeky: I think the default behaviour (prefixing an asterisk to required fields) also adds an extra space. I've fixed this in this instance however.

Also, same with renaming parameter to `optional`.

app/controllers/addresses_controller.rb
```diff
@@ -1,18 +1,20 @@
 class AddressesController < ApplicationController
   def create
-    if property_reference.blank?
-      return redirect_to(
-        address_search_path,
-        alert: t('addresses.errors.blank')
-      )
-    end
+    @form = AddressForm.new(address_form_params)
```
dgmstuart, line 3: Nice - I like this :)

app/controllers/addresses_controller.rb
```diff
@@ -21,11 +23,7 @@ def create
 
   private
 
-  def update_params
-    params.require(:address).permit(:property_reference)
-  end
-
-  def property_reference
-    update_params[:property_reference]
+  def address_form_params
```
dgmstuart, line 26: Ah this tidy up feels good

dgmstuart (approved):
Looks great to me 👍

PR #20, opened 2017-10-09
(93) Display selected answers on confirmation

dgmstuart:
Previously we had a hardcoded confirmation page. Now use the Confirmation presenter object to get the answers and format them appropriately.

Of course this requires us to have actually stored the answers. In some cases this is simple (e.g. description has one field and no validations yet). In other cases it's fiddly (e.g. contact details has multiple fields and validations) so we delegate the saving of the answer to another object.


spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
```diff
@@ -2,8 +2,38 @@
 
 RSpec.feature 'Resident can see a confirmation of their repair request' do
   scenario 'when the repair request creation was successful' do
-    visit '/confirmations/00012143'
+    property = {
+      'property_reference' => '00000503',
+      'short_address' => 'Ross Court 30',
+      'postcode' => 'E5 8TE',
```
leeky: Is the API `list_properties` endpoint being changed to return the postcode in the results, because it doesn't at present..?

spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
```diff
@@ -28,12 +29,22 @@
...
 
     expect(page).to have_content 'Your reference number is abc123'
+
+    expect(page).to have_content "Description: The bath won't drain"
+
+    expect(page).to have_content 'Name: John Smith'
+    expect(page).to have_content 'Phone: 01234 567 890'
     expect(page).to have_content 'Address: Ross Court 30, E5 8TE'
+
+    expect(page).to have_content 'Time: afternoon (12pm - 5pm)'
```
leeky: I think we should have a test for what happens when both callback times are checked, as I'm not sure what the output should look like.

app/views/confirmations/show.html.haml
```diff
@@ -31,15 +31,13 @@
     %h2 Summary
 
     %h3 Problem
-    %p Description: The bath won't drain
+    %p= t('confirmation.summary.description', description: "The bath won't drain")
```
leeky: :+1:

app/views/confirmations/show.html.haml
```diff
@@ -31,15 +31,13 @@
...
-    %p
-      Address:
-      = @confirmation.address
+    %p= t('confirmation.summary.name', name: 'John Smith')
+    %p= t('confirmation.summary.phone', phone: '01234 567 890')
+    %p= t('confirmation.summary.address', address: @confirmation.address)
 
     %h3 Callback time
-    %p Time: afternoon (12pm - 5pm)
+    %p= t('confirmation.summary.time', time: 'afternoon (12pm - 5pm)')
```
leeky: Is it worth renaming this to `callback_time`, as it might get confusing when we start dealing with `appointment_time` on the same page(?)

app/controllers/describe_repair_controller.rb
```diff
@@ -1,7 +1,17 @@
 class DescribeRepairController < ApplicationController
-  def index; end
+  def index
+    @form = DescriptionForm.new
```
leeky: Not sure if it matters, but I've been trying to keep the form and controller names in sync - so `DescribeRepairController` would reference `DescribeRepairForm`.

spec/models/description_form_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+require 'app/models/description_form'
+
+RSpec.describe DescriptionForm do
+  describe 'validations' do
+    it 'is valid without a description' do
+      expect(DescriptionForm.new.valid?).to eq true
+    end
+
+    it 'is valid with a description' do
+      expect(DescriptionForm.new(description: 'my taps are leaking').valid?).to eq true
```
leeky: How about splitting this into two lines to improve readability? i.e.

```ruby
form = DescriptionForm.new(description: 'my taps are leaking')
expect(form).valid?).to eq true
```

(NB: Because of RSpec's built-in matchers, you can also do `expect(form).to be_valid`)

leeky (changes requested):

spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
dgmstuart: Yes: https://trello.com/c/VNIlJMI7/113-the-properties-propertyref-endpoint-should-return-postcode

I'm assuming this endpoint will ultimately return significantly more data - it was only this minimal as a first stab.

I considered merging the postcode into the answer hash ourselves, but we'd have to pass it from the radio buttons, which could get quite funky.

spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
dgmstuart: Yup - that's covered later in the `Confirmation` presenter object: the approach I took was to test specifics in that object, and then have just one feature spec which tests the integration.

app/views/confirmations/show.html.haml
dgmstuart: 👍  Good shout - will do

app/controllers/describe_repair_controller.rb
dgmstuart: Sure, that makes sense 👍

dgmstuart:
Will pick up these tasks tomorrow

spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
leeky: :+1:

I did spot this after I posted my previous comment, but then forgot to go back and remove it!

dgmstuart:
@leeky I think I've addressed all the comments - ready for another look

dgmstuart:
DON'T MERGE YET - spotted that the callback time isn't displayed properly in the body text

leeky (approved):

PR #21, opened 2017-10-10
(96) Users can continue if no addresses found

dgmstuart:
This situation is essentially the same as if there were addresses found,
but none of them are where the user lives.

I initially tried making the button a link, which seemed simpler, but
that means that the redirection path (to the "we cannot find your
address" page) is duplicated in two places: the controller and the view.

Submitting a form with a hidden field feels a bit clunkier, but on the
plus side, both scenarios (no addresses at all, user's address not
shown) are handled by the same code.

Ensure that Postcode is sent in both scenarios. It's only actually used
in one specific case, but it feels more predictable if the view always
returns the same data.

### DIFF
--- a/app/views/address_searches/create.html.haml
+++ b/app/views/address_searches/create.html.haml
@@ -13,9 +13,10 @@
 
 - if @address_search_results
   #address-search-results
-    - if @address_search_results.any?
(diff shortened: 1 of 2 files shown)


leeky (approved):
LGTM! :shipit:

PR #22, opened 2017-10-10
(125) Create a repair in UH

dgmstuart:
Display the resulting repair reference on the confirmation page(125) Create a repair in UH


leeky (approved):
LGTM!

PR #23, opened 2017-10-11
(120) User can diagnose their problem by answering a series of questions

leeky:
Implement a decision tree which is read from a YAML file.


spec/models/question_set_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        },
+      }
+
+      expect(store.questions).to eql expected
+    end
+  end
+
+  describe '#find' do
+    it 'returns the requested question' do
+      store = QuestionSet.new('spec/fixtures/basic_question.yml')
```
dgmstuart, line 20: Nice approach. Gives a lot of confidence. I thought they might be slow since they're hitting the disk, but they're 10x faster than the feature specs.

spec/models/question_set_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+          'question' => 'Is this an example?',
+          'answers' => [
+            { 'text' => 'Definitely' },
+            { 'text' => 'Probably' },
+          ],
+        },
+      }
+
+      expect(store.questions).to eql expected
+    end
```
dgmstuart: There are no tests for how this can go wrong, but maybe we don't need to worry yet?

I guess if the file isn't YAML we'll get a YAML error, which is fine. Not sure what the implications of the structure being wrong would be (e.g. if `answers` was not an array).

app/models/question_set.rb
```diff
@@ -14,7 +14,7 @@ def find(id)
       raise UnknownQuestion, "Cannot find question with id '#{id}'"
     end
 
-    questions.fetch(id)
+    Question.new(questions.fetch(id))
```
dgmstuart: If we created these objects on initialization rather than on `find`, would that mean that we caught some possible mistakes in the YAML at boot time rather than at runtime? Would it be worth it?

app/models/question_set.rb
```diff
@@ -0,0 +1,20 @@
+class QuestionSet
+  class UnknownQuestion < StandardError; end
+
+  attr_reader :questions
```
dgmstuart: Is this only exposed for the purposes of testing?

spec/features/users_can_diagnose_their_issue_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.feature 'Users can diagnose their issue' do
+  scenario 'viewing the first question' do
+    allow_any_instance_of(QuestionSet)
+      .to receive(:questions)
```
dgmstuart: The method which our application code is calling is `find` - would it make more sense to stub that instead?

Is this currently relying on the fact that we know that `find` is defined in terms of `questions`?

spec/models/question_set_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+RSpec.describe QuestionSet do
+  describe '#initialize' do
```
dgmstuart, line 4: Is this actually a test of the `#questions` method rather than intialization?

spec/models/question_set_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        'example' => {
+          'question' => 'Is this an example?',
+          'answers' => [
+            { 'text' => 'Definitely' },
+            { 'text' => 'Probably' },
+          ],
+        },
+      }
+
+      expect(store.questions).to eql expected
```
dgmstuart: Is this test redundant? Do we care how `QuestionSet` internally represents the data? If not then is the test of the #find method sufficient?

spec/models/question_set_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        'answers' => [
+          { 'text' => 'Definitely' },
+          { 'text' => 'Probably' },
+        ],
+      }
+
+      expect(store.find('example')).to eql expected
+    end
+
+    it 'raises an error if the requestion question was not found' do
```
dgmstuart: Typo "requestion question" 

(maybe only change if you touch this file for other reasons?)

spec/features/users_can_diagnose_their_issue_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.feature 'Users can diagnose their issue' do
+  scenario 'viewing the first question' do
+    allow_any_instance_of(QuestionSet)
```
dgmstuart: `allow_any_instance_of` is intended for use with legacy code and the RSpec team recommend avoiding it if possible: https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance

Would the approach we took for `JsonApi` work here? e.g. in spec/features/residents_can_see_confirmation_of_repair_request_spec.rb

spec/features/users_can_diagnose_their_issue_spec.rb
```diff
@@ -20,4 +20,56 @@
...
+    visit '/questions/first'
+    choose_radio_button 'Yes'
+    click_on 'Continue'
+
+    expect(page.current_path).to eql '/questions/second'
+    expect(page).to have_content 'Are you sure?'
+    expect(page).to have_unchecked_field 'Absolutely!'
+  end
+
+  scenario 'choosing an answer redirects to the address search page by default' do
```
dgmstuart: The title says "address search page" - should this be "problem description"?

app/models/question.rb
```diff
@@ -10,5 +10,15 @@ def initialize(question_hash)
...
     answers.map { |answer| answer['text'] }
   end
+
+  def redirect_path_for_answer(chosen_answer)
+    answer_hash = answers.detect { |answer| answer['text'] == chosen_answer }
+
+    if answer_hash.key?('goto')
+      Rails.application.routes.url_helpers.questions_path(answer_hash['goto'])
+    else
+      Rails.application.routes.url_helpers.describe_repair_path
```
dgmstuart: Did you consider putting the `Rails.application.routes.url_helpers` in the initializer?

spec/models/question_spec.rb
```diff
@@ -32,4 +32,20 @@
...
+        'question' => 'Where do you want to go?',
+        'answers' => [
+          {
+            'text' => 'London',
+            'goto' => 'next',
+          },
+        ],
+      )
+
+      expect(question.redirect_path_for_answer('London')).to eql '/questions/next'
```
dgmstuart, line 48: Is this now an integration test since it's testing the routing layer as well?

app/models/question.rb
```diff
@@ -10,5 +10,15 @@ def initialize(question_hash)
   def answers_for_collection
     answers.map { |answer| answer['text'] }
   end
+
+  def redirect_path_for_answer(chosen_answer)
+    answer_hash = answers.detect { |answer| answer['text'] == chosen_answer }
+
+    if answer_hash.key?('goto')
+      Rails.application.routes.url_helpers.questions_path(answer_hash['goto'])
```
dgmstuart: Is `goto` always referring to a question ID? If so is there a more descriptive name we could give it?

app/views/pages/toilet_unblock_info.html.haml
```diff
@@ -0,0 +1,8 @@
+%h1 Toilet blocked?
+
+%p It's pretty straightforward to unblock a toilet, you'll need:
+
+%ul
+  %li A large bin bag
+  %li A mop
```
dgmstuart, line 7: Do we have/need something in Trello to make sure we fill these in with real content at some point?

db/questions.yml
```diff
@@ -35,43 +35,155 @@ kitchen_problem:
...
+is_bath_blocked_or_cracked:
+  question: Is your bath blocked or cracked?
+  answers:
+    - text: 'Yes'
+    - text: 'No'
+
+bathroom_lights_problem:
+  question: Is the problem with the...
+  answers:
+    - text: light fitting
```
dgmstuart: Should we standardise on Capitalised text?

db/questions.yml
```diff
@@ -129,29 +244,52 @@ fire_door:
...
 toilet_problem:
-  question: What's wrong with your toilet?
+  question: Is your toilet...?
   answers:
-    - text: Blocked or not flushing
+    - text: blocked
       goto: blocked_only_toilet
 
-    - text: It's cracked
+    - text: not flushing
```
dgmstuart: Capitalise these?

dgmstuart (changes requested):
Good job

A few comments and questions

app/models/question.rb
leeky: I mainly chose it because it has the same number of characters as `text` - so it's possibly easier to read when you're editing the file as a non-developer.

leeky: As discussed, I've changed `goto` to `next`.

app/views/pages/toilet_unblock_info.html.haml
leeky: I've added a card to the backlog for these - https://trello.com/c/OyIiRYa3/146-content-for-static-pages

db/questions.yml
leeky: :+1:

spec/features/users_can_diagnose_their_issue_spec.rb
leeky: We discussed a solution, and I've push a commit which switches to an `instance_double`-based stub 👍

spec/models/question_set_spec.rb
leeky: I've added [some tests](92b58d5ea8be928127cb27786e230be7c7f5e753) that do basic YAML validation.

spec/models/question_set_spec.rb
leeky: :+1:

leeky:
@dgmstuart This should be ready for another look.

dgmstuart (approved):
LGTM - Great job!

PR #25, opened 2017-10-16
(130) UI tweaks

dgmstuart:
Merge #23 first


app/assets/stylesheets/components/main.scss
```diff
@@ -30,6 +30,10 @@ main {
   ul {
     list-style-type: disc;
     padding-left: 20px;
+
+    li {
+      margin-bottom: 10px;
```
leeky: I tend to use `em` measurements for this sort of thing, so that the spacing is relative to the size of the text it's applied to. 

Maybe replace with with `0.5em` ?

app/views/pages/start.html.haml
```diff
@@ -10,15 +10,16 @@
 %p
   Reporting a repair should take around 5 minutes.
 
+%p.notice
```
leeky, line 13: Nice change :+1:

leeky (changes requested):

dgmstuart:
@leeky I've amended the original commit to use em - ready for another look

leeky (approved):

leeky:
LGTM! :shipit:

PR #26, opened 2017-10-17
(138) Form submit buttons are always shown

dgmstuart:
We thought this was a GDS-ish pattern, but actually it's not, and leads
to complications e.g. when revisiting a form (an option is already
selected, but the button is not visible)


spec/features/resident_can_locate_problem_spec.rb
```diff
@@ -100,7 +100,7 @@
     expect(page).to have_content('N1 6NU')
   end
 
-  scenario 'performing a search and not selecting an address, prevents form submission', js: true do
+  scenario 'performing a search and not selecting an address, prevents form submission' do
```
leeky: I think this test can be removed altogether - the non-JS path is tested by `performing a search and not selecting an address` below.

leeky (changes requested):

spec/features/resident_can_locate_problem_spec.rb
dgmstuart: Ah - sorry: I was too blasé and didn't actually read the test. Amended the original commit

dgmstuart:
@leeky ready for another look

leeky (approved):

leeky:
LGTM! :shipit:

PR #27, opened 2017-10-17
(127) Repairs cannot be submitted to the API without a description

leeky:
The API requires the `problem` attribute to be set when creating a repair, so set it to `n/a` if the user didn't complete the optional 'describe repair' form.


dgmstuart (approved):
LGTM - hopefully we'll pass something meaningful in this before too long, but this fixes the problem for now.

PR #30, opened 2017-10-18
(153) Start page content

leeky:
start-page-new


dgmstuart (changes requested):
I hate to be that guy, but does this maybe "need more whitespace"? On the example on gov.uk the page is broken up into 3 distinct sections:
https://www.gov.uk/service-manual/design/start-pages

At the moment the start page feels like all one block, which might feel intimidating.

Perhaps some more space between title, "what this is" and "before you start" would help with that?

Also in the "before you start" section the spacing between the title and the two paras feels odd to me. In the Gov.uk example these are evenly spaced.

leeky:
I've given the styling a good kick - increasing whitespace, using the correct font etc, and added some commits to this branch:

report-a-repair

I've tested in with an older iPhone with BrowserStack, and it seems happier on there too.

report-a-repair-mobile

leeky:
@dgmstuart Should be ready for another look.

dgmstuart (approved):

PR #32, opened 2017-10-18
(139) Send SOR codes to UH

dgmstuart:
if we have an SOR code, send it to UH, if not send just the repair request parts

To test, add a valid SOR code into a leaf node of the yaml file

### DIFF
--- a/app/models/question.rb
+++ b/app/models/question.rb
@@ -2,7 +2,6 @@ class Question
   class InvalidAnswerError < StandardError; end
 
   attr_reader :title
-  attr_reader :answers
   attr_reader :next_question
 
   def initialize(question_hash)
@@ -12,17 +11,17 @@ class Question
   end
 
   def answers_for_collection
-    answers.map { |answer| answer['text'] }
+    @answers.map { |answer| answer['text'] }
   end
 
   def answer_data(chosen_answer)
-    answers.detect { |answer| answer['text'] == chosen_answer }.tap do |data|
+    @answers.detect { |answer| answer['text'] == chosen_answer }.tap do |data|
       raise InvalidAnswerError if data.nil?
     end
   end
 
   def redirect_path_for_answer(chosen_answer)
-    answer_hash = answers.detect { |answer| answer['text'] == chosen_answer }
+    answer_hash = answer_data(chosen_answer)
 
     if answer_hash.key?('next')
       Rails.application.routes.url_helpers.questions_path(answer_hash['next'])
@@ -42,6 +41,6 @@ class Question
   end
 
   def multiple_choice?
-    answers.any?
+    @answers.any?
   end
 end
--- a/app/models/question.rb
+++ b/app/models/question.rb
@@ -1,4 +1,6 @@
 class Question
+  class InvalidAnswerError < StandardError; end
+
   attr_reader :title
   attr_reader :answers
   attr_reader :next_question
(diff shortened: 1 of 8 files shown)


app/services/create_repair.rb
```diff
@@ -4,22 +4,33 @@ def initialize(api: HackneyApi.new)
...
+    }.tap do |hash|
+      hash[:repairOrders] = create_work_order_params(params) if params.sor_code
+    end
+  end
+
+  def create_work_order_params(params)
+    [
+      {
+        jobCode: params.sor_code,
+        propertyReference: params.property_reference,
```
leeky, line 28: I thought we were going to change the API endpoint to remove the duplication of `propertyReference` ?

Also, it's referred to as both `propertyRef` and `propertyReference` in the same hash, which feels odd.

app/services/create_repair.rb
```diff
@@ -35,6 +46,10 @@ def property_reference
     def priority
       'N'
     end
+
+    def sor_code
+      @answers.fetch('diagnosis', {})['sor_code']
```
leeky, line 54: Should this be `@answers.fetch('diagnosis', {}).fetch('sor_code')`, so avoid a `NilClass` error if `diagnosis` is nil..?

app/models/question.rb
```diff
@@ -13,6 +15,12 @@ def answers_for_collection
...
   end
 
+  def answer_data(chosen_answer)
+    answers.detect { |answer| answer['text'] == chosen_answer }.tap do |data|
+      raise InvalidAnswerError if data.nil?
+    end
+  end
+
   def redirect_path_for_answer(chosen_answer)
     answer_hash = answers.detect { |answer| answer['text'] == chosen_answer }
```
leeky: If it worth using the  new `answer_data` method here?

leeky (changes requested):

app/models/question.rb
leeky: As discussed, you did this in the next commit!

app/services/create_repair.rb
dgmstuart: @leeky in what circumstance could we get a NilClass error?
I think all the possibilities are covered by tests - here are the specific cases:

```
> @answers = {}
> @answers.fetch('diagnosis', {})['sor_code']
=> nil
                              
> @answers = { 'diagnosis' => { 'next' => 'foo' }}
> @answers.fetch('diagnosis', {})['sor_code']
=> nil

> @answers = { 'diagnosis' => { 'sor_code' => 'foo' }}
> @answers.fetch('diagnosis', {})['sor_code']
=> "foo"
```

app/services/create_repair.rb
dgmstuart: Totally agree - I should have mentioned in the commit message: these values work with the Hackney API as it is currently deployed. When https://trello.com/c/txl3rMJK/132-apply-structure-to-endpoints is done we can change it to something more sensible.

app/services/create_repair.rb
dgmstuart: For completeness:

```
> @answers = { 'diagnosis' => { 'next' => 'foo' }}
> @answers.fetch('diagnosis', {}).fetch('sor_code')
KeyError: key not found: "sor_code"
```

app/services/create_repair.rb
leeky: 👍

leeky (approved):

leeky:
LGTM! :shipit:

app/services/create_repair.rb
leeky: My mistake - I'd forgotten that requesting the value of a missing key in a `Hash` returns `nil`.

PR #34, opened 2017-10-18
(163) fix bugs in CreateRepair

dgmstuart:
A couple of niggles appeared after rebasing and merging (139) Send SOR codes to UH


leeky (approved):
LGTM! :shipit:

PR #35, opened 2017-10-18
(164) Ignore ABC complexity rubocop fails

dgmstuart:
I think all of these will be dealt with in due course, but for now we
need to get CI running.(chore) Ignore ABC complexity rubocop fails


leeky (approved):
LGTM! :shipit:

PR #36, opened 2017-10-20
(165) Diagnosed repairs don't involve a callback

dgmstuart:
Basic approach: rename the existing controller to "ContactDetailsWithCallback" and create a new one for the diagnosed case.

There's a lot of refactoring required here: To extract the common things between the two contact details controllers, and to make the controllers smaller since the new changes make them unacceptably long. I would be happy to revert the latter if they prove controversial.


dgmstuart:
Created from the wrong branch by accident

PR #37, opened 2017-10-23
(165) Diagnosed repairs don't involve a callback

dgmstuart:
Basic approach: rename the existing controller to "ContactDetailsWithCallback" and create a new one for the diagnosed case.

There's a lot of refactoring required here: To extract the common things between the two contact details controllers, and to make the controllers smaller since the new changes make them unacceptably long. I would be happy to revert the latter if they prove controversial.


app/services/create_repair.rb
```diff
@@ -40,5 +40,9 @@ def request_reference
         @result.fetch('repairRequestReference')
       end
     end
+
+    def work_order_reference
+      @result.fetch('workOrderReference')
```
dgmstuart: Oops - although this should ultimately be what this field is called, currently it's called just `orderReference` in the response from the API

app/presenters/callback.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(time_slot:, request_reference:)
+    @time_slot = time_slot
+    @request_reference = request_reference
+  end
+
+  def time
+    case @time_slot
+    when ['morning']
+      'between 8am and 12pm'
```
leeky: The view prefixes all of these with 'between', so that isn't needed here.

app/views/confirmations/_callback.html.haml
```diff
@@ -0,0 +1,14 @@
+%p
+  We will call you within one working day, between
+  %strong= callback.time
```
dgmstuart, line 3: [Alex has pointed out]( https://trello.com/c/e1aGvtgs/166-content-tweak) that the callback time also includes the word "between" - perhaps we should fix that repetition in this PR?

app/controllers/contact_details_controller.rb
```diff
@@ -11,8 +11,10 @@ def submit
     contact_details_saver =
       ContactDetailsSaver.new(selected_answer_store: selected_answer_store)
     contact_details_saver.save(@form)
+    result =
+      CreateRepair.new.call(answers: selected_answer_store.selected_answers)
 
-    redirect_to confirmation_path('WORK_ORDER_REFERENCE')
+    redirect_to confirmation_path(result.work_order_reference)
```
leeky: Is it possible for the `work_order_reference` used here and `request_reference` used in `ContactDetailsWithCallbackController` to clash?

spec/support/capybara_helpers.rb
```diff
@@ -2,4 +2,13 @@ module CapybaraHelpers
   def choose_radio_button(label)
     find(:label, text: label).click
   end
+
+  def stub_one_diagnosis_question(answers:, question: 'Dummy question', id: 'location')
```
leeky: As it's `question` in the singular, could this be called `stub_diagnosis_question` ?

app/models/address_search_form.rb
```diff
@@ -1,4 +1,4 @@
-class AddressSearch
+class AddressSearchForm
```
leeky, line 1: :+1:

app/services/callback_time_saver.rb
```diff
@@ -0,0 +1,23 @@
+class CallbackTimeSaver
+  def initialize(selected_answer_store:)
```
leeky, line 2: As with the change to `address_saver.rb` later on, is it worth defaulting `selected_answer_store` to `SelectedAnswerStore` so we don't have to keep new-ing this up everywhere?

app/presenters/appointment.rb
```diff
@@ -0,0 +1,5 @@
+class Appointment
+  def to_partial_path
+    '/confirmations/appointment'
```
leeky, line 3: I think having this point at an empty template for now is fine, since we'll soon want to add content here.

app/views/confirmations/_callback.html.haml
leeky: Seems sensible to me :+1:

spec/support/capybara_helpers.rb
leeky: This change makes everything a lot tidier :smile:

leeky (changes requested):

app/services/callback_time_saver.rb
dgmstuart: Yeah, that would be preferable - the problem is that `SelectedAnswerStore` needs the `session` and I don't think that's accessible outside the Controller.

app/presenters/callback.rb
dgmstuart: 👍 
My thought was originally that this could support different formats of slot e.g. `before 5pm`, but in retrospect I think that's total YAGNI.

app/controllers/contact_details_controller.rb
dgmstuart: UGH - you're totally right: the confirmation page will need to know which thing to request. This needs a rethink.

spec/support/capybara_helpers.rb
dgmstuart: Renamed

app/services/create_repair.rb
dgmstuart: Renamed

app/views/confirmations/_callback.html.haml
dgmstuart: Done

dgmstuart:
@leeky Added some commits addressing your comments. 

Passing a sensible ID to the confirmations page now involves making an additional request to fetch the repair data, but I think it's a request we would be making eventually anyway

Should be ready for another look.

app/services/callback_time_saver.rb
leeky: :+1:

leeky (approved):

leeky:
LGTM! :shipit:

PR #38, opened 2017-10-23
(168) Updated questions & workflows

JasonEwins:
Significant changes to existing repairs workflows

To consider:
- Created multiple instances of 'describe_item_problem pages 
Do we need separate instance of page when a simple text change is required?

- Created 3 individual instances of electrical_problems for each room type
Is there a way to use one global electrical_problems branch and then +- repair types per room?

- Gas & Carbon Monoxide need to divert to national grid contact message (new page)

- Require additional pages for light bulb check question & resulting help messaging

### DIFF
--- a/db/questions.yml
+++ b/db/questions.yml
@@ -20,9 +20,9 @@ describe_communal_problem:
     - text: The main entrance door is damaged or not closing
     - text: The intercom is not working for my property
     - text: The lift is faulty
-    - text: There are pests or rodents
-    - text: There is a blocked drain or gully
     - text: The rubbish chute is blocked or damaged
+    - text: There is a blocked drain or gully
+    - text: There are pests or rodents
     - text: Something else
 
 faulty_lamp_post:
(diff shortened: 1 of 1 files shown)


db/questions.yml
```diff
@@ -59,103 +59,104 @@ which_room:
...
+      next: electrical_hazard_warning
+
+    - text: Smoke detector is beeping
+      page: emergency_contact
 
     - text: Sockets
       page: electrical_hazard_warning
 
-other_electrical_problem:
+    - Other: kitchen_other_problem
```
leeky: I'm guessing you were meant to write the following:

```ruby
  - text: Other
    goto: kitchen_other_problem
```

db/questions.yml
```diff
@@ -59,103 +59,104 @@ which_room:
...
-      next: is_basin_leaking_or_cracked
-
-    - text: Extractor fan
-      next: is_fan_leaking_water
+    - text: Electrical
+      next: bathroom_electrical_problem
 
-    - text: Lights
-      next: bathroom_lights_problem
+    - next: Heating
```
leeky: This should be `text: Heating`.

db/questions.yml
```diff
@@ -164,115 +165,125 @@ is_fan_leaking_water:
...
 
-    - text: 'No'
-      next: window_material
+    - text: Carbon Monoxide detector is beeping
+      page: emergency_contact  
+
+    - text: Sockets
+      page: electrical_hazard_warning
+
+    - Other: other_electrical_problem
```
leeky: And again here.

leeky (changes requested):
Spotted a few typos in this, which would cause the app to stop working.

db/questions.yml
JasonEwins: Perfect, thanks.

db/questions.yml
JasonEwins: Ah, I see - should this be the same for each instance of 'Other'..? would you like me to amend these or you happy to update?

leeky (approved):

PR #40, opened 2017-10-24
(90) Include Hotjar and Google analytics code

dgmstuart:
Don't show anything unless environment variables are present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,41 @@
-# Hackney - Maintaining my Home
+# Hackney - Report a Repair
 
-This README would normally document whatever steps are necessary to get the
-application up and running.
+## Configuration
 
-Things you may want to cover:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,16 +1,35 @@
-# Hackney - Maintaining my Home
+# Hackney - Report a Repair
```
leeky, line 1: :+1:

leeky (approved):
I booted the app up with some dummy environment variables, and the embed codes were added as expected, so LGTM! :shipit:

leeky:
As discussed, updated `README.md` to mention `.env.local` rather than `.env`, as the former doesn't get loaded in the test environment. Also added the environment variables used to enable HTTP Basic Authentication.

PR #41, opened 2017-10-25
(144) 'Tell us more' page: customisable help text

dgmstuart:
- Use a 'desc:' key with the name of a partial
- Describe form is required if no SOR code
  - GOTCHA: clear your session if you have ever sent a request with an SOR code: this will be hanging around in your session cookie
- Removed the parts of the Question code which we didn't end up using


app/controllers/application_controller.rb
```diff
@@ -2,4 +2,8 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   include Authentication
+
+  def selected_answer_store
+    @selected_answer_store ||= SelectedAnswerStore.new(session)
```
leeky, line 7: Whilst it does make `selected_answer_store` somewhat of a mystery guest,  it does tidy things up quite a bit

👍

app/presenters/describe_repair.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def form
+    @form ||= build_form(@form_params)
+  end
+
+  private
+
+  def diagnosed?
+    !RepairParams.new(@answers).sor_code.nil?
```
leeky: How about this instead:

`RepairParams.new(@answers).sor_code.present?`

spec/fixtures/missing_partials.yml
```diff
@@ -0,0 +1,7 @@
+---
```
leeky, line 1: I think this file belongs in the next commit(?)

spec/rails_helper.rb
```diff
@@ -2,6 +2,7 @@
 SimpleCov.start 'rails' do
   add_group 'Queries', 'app/queries'
   add_group 'Validators', 'app/validators'
+  minimum_coverage 98
```
leeky: Should we add `add_group 'Presenters', 'app/presenters'` for completeness?

app/controllers/describe_unknown_repair_controller.rb
```diff
@@ -1,18 +1,25 @@
 class DescribeUnknownRepairController < ApplicationController
   def index
+    @back = Back.new(controller_name: 'questions/start')
```
leeky: As discussed, this `Back` shouldn't be here.

app/controllers/describe_unknown_repair_controller.rb
```diff
@@ -1,18 +1,25 @@
...
     @form = DescribeUnknownRepairForm.new
+    @partial = 'describe_problem'
+
+    render 'describe_repair/index'
   end
 
   def submit
-    @form = DescribeUnknownRepairForm.new(describe_unknown_repair_form_params)
-    return render :index unless @form.valid?
+    @back = Back.new(controller_name: 'questions/start')
```
leeky: And this one too.

leeky (changes requested):
Just a few minor comments, otherwise nice work :tada:

app/presenters/describe_repair.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_reader :form
+
+  def initialize(form_params: {}, details:, answers:, partial_checker:)
+    @form_params = form_params
+    @answers = answers
+    @details = details
+    @partial_checker = partial_checker
+  end
+
+  def partial
```
leeky, line 11: Just playing with this through the frontend, and I was wondering whether this should be...

```ruby
  def partial
    return @details if @partial_checker.exists?(@details)

    if diagnosed?
      'anything_else'
    else
      'describe_problem'
    end
  end
```

i.e. Always use the partial referenced in the `desc` key, but failover to `anything_else` or `describe_problem` depending on whether we have a SOR code or not(?)

I spotted this as I inadvertently had a `sor_code` stored in my session, so it behaved as if the `?details` parameters was being ignored, as `diagnosed?` was always true.

app/controllers/describe_unknown_repair_controller.rb
dgmstuart: I have eliminated it from history like a time travel B-movie

app/presenters/describe_repair.rb
dgmstuart: That is indeed clearer. I thought this wouldn't be available in this class, but I guess ActiveModel includes it. 

Since we're asking this question of RepairParams in two places, I've created a `diagnosed?` method: that also makes us more flexible if we wanted to redefine what it means to be 'diagnosed` in future.

spec/fixtures/missing_partials.yml
dgmstuart: OOPS!!!!

dgmstuart: Fixed

spec/rails_helper.rb
dgmstuart: Yes! good point

app/controllers/addresses_controller.rb
```diff
@@ -13,7 +13,7 @@ def create
       return render 'address_searches/create'
     end
 
-    if RepairParams.new(selected_answer_store.selected_answers).sor_code
```
leeky, line 17: Nice addition :+1:

app/presenters/describe_repair.rb
dgmstuart: Agreed - good fix. The problem will still be that the form should be required, but will be optional, but that's a different problem - fixed by resetting the session.

dgmstuart:
@leeky ready for another look

leeky (approved):

leeky:
LGTM! :shipit:

PR #42, opened 2017-10-30
(161) Residents can book a repair appointment

leeky:
Allow a resident to choose from a list of available appointments. 

NB: Whilst the tests pass, the API doesn't yet provide the real endpoints, so this cannot be run 'live'. Also, the stubbed endpoints are subject to change.

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -1,0 +1,11 @@
+class AppointmentsController < ApplicationController
+  def show
+    @appointments = Appointments.new(['Thursday Afternoon (11th October)'])
+  end
+
+  def submit
+    redirect_to confirmation_path(params[:work_order_reference])
+  end
+
+  Appointments = Struct.new(:slots_for_collection)
+end
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -12,11 +12,7 @@ class AppointmentsController < ApplicationController
       return render :show
     end
 
-    HackneyApi.new.book_appointment(
-      work_order_reference: work_order_reference,
-      begin_date: @form.begin_date,
-      end_date: @form.end_date
-    )
+    book_appointment_save_into_answer_store
 
     redirect_to confirmation_path(params[:repair_request_reference])
   end
@@ -41,5 +37,17 @@ class AppointmentsController < ApplicationController
     )
 
     appointments.map { |a| AppointmentPresenter.new(a) }
+  end
+
+  def book_appointment_save_into_answer_store
+    appointment = HackneyApi.new.book_appointment(
+      work_order_reference: work_order_reference,
+      begin_date: @form.begin_date,
+      end_date: @form.end_date
+    )
+
+    selected_answer_store.store_selected_answers(
+      'appointment', appointment
+    )
   end
 end
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -12,7 +12,11 @@ class AppointmentsController < ApplicationController
       return render :show
     end
 
-    # TODO: Book appointment
+    HackneyApi.new.book_appointment(
+      work_order_reference: work_order_reference,
+      begin_date: @form.begin_date,
+      end_date: @form.end_date
+    )
 
     redirect_to confirmation_path(params[:repair_request_reference])
   end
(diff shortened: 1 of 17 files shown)


app/presenters/appointment_presenter.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def appointment_id
+    [
+      @appointment.fetch('beginDate'),
+      @appointment.fetch('endDate'),
+    ].join('|')
+  end
+
+  def description
+    "#{day_of_week} #{begin_time}-#{end_time} (#{day} #{month})"
```
dgmstuart, line 16: Nicely broken out 👍

app/controllers/appointments_controller.rb
```diff
@@ -1,11 +1,41 @@
...
+    redirect_to confirmation_path(params[:repair_request_reference])
+  end
+
+  private
+
+  def appointment_form_params
+    params.require(:appointment_form).permit(:appointment_id)
+  end
+
+  def work_order_reference
```
dgmstuart, line 26: The original idea was that `Appointments` would be a facade exposing `form` and `slots_for_collection` and providing a place for these methods to live so that we don't have to add them in the controller.

This approach works for now but would it be more flexible to move these methods into an object?

app/models/appointment_form.rb
```diff
@@ -8,4 +8,18 @@ class AppointmentForm
...
+  end
+
+  def end_date
+    split_date.last
+  end
+
+  private
+
+  def split_date
+    @split_date ||= @appointment_id.to_s.split('|')
```
dgmstuart, line 27: Is this because the ID will be of form `firstdate|lastdate`? Does this need to be shown in a test somewhere?

app/presenters/appointment_presenter.rb
```diff
@@ -51,10 +51,20 @@ def day
...
-    I18n.l(begin_date, format: '%l:%M%P').strip
+    friendly_time(begin_date)
   end
 
   def end_time
-    I18n.l(end_date, format: '%l:%M%P').strip
+    friendly_time(end_date)
+  end
+
+  def friendly_time(time)
```
dgmstuart: Could this be extracted to an initializer? e.g. https://robots.thoughtbot.com/custom-formats-for-datetime

spec/presenters/appointment_presenter_spec.rb
```diff
@@ -49,7 +59,7 @@
         'bestSlot' => true
       )
 
-      expect(presenter.time).to eql '12:00pm and 4:45pm'
+      expect(presenter.time).to eql '12pm and 4:45pm'
```
dgmstuart: You've handled this time display nicely 👍

dgmstuart (dismissed):
Can't test from over here, but looks good - good job 👍

app/models/appointment_form.rb
leeky: Yeah, this exists because appointments don't have IDs - in fact, we still have no idea whether start/end time is enough to book with - I will add some tests around this.

app/presenters/appointment_presenter.rb
leeky: I thought about this, but I've read the GOVUK guidelines that suggest using "midday" in place of "12pm", which seems like a good idea - and this wouldn't work with Rails' standard datetime formatting.

I'll add a card to add the 'midday' special case in, and leave this method as-is for now(?)

app/controllers/appointments_controller.rb
leeky: I'm going to leave this in the controller for now I think. If it gets any more complex, I'll revisit and extract this into smaller objects.

leeky:
@dgmstuart Should be ready for another look.

techbelly:
This looks ok to me. One concern which isn't enough for me to reject it.

If I've understood it correctly, I'm a tiny bit nervous that we present the user a list of appointments, and they could pass us back a start time and end time that might not be in that list. That's nice and stateless, which is good. But it's open to user tampering.

There are a couple of simple ways around this - we could cryptographically sign the ids, I think there's some support in rails for that. We could also store the appointment list in the session and index into it.

I'm not sure I'm worried about this in this case. Up to you if you want to take that further.

techbelly (approved):
Added a comment on general approach to ids. Code looks good to me, and I understand your responses to duncan's questions.

PR #43, opened 2017-10-30
Revised workflows - SOR review 30/10

JasonEwins:
Revised the following workflows following review of SOR

Heating: 
- boiler issues: in-dwelling & communal
- Removed electric heating repairs

window:  
- material selection
- Revised glazing repair

external doors: 
- material selection
- Door locks

### DIFF
--- a/db/questions.yml
+++ b/db/questions.yml
@@ -103,11 +103,9 @@ kitchen_electrical_problem:
 heating_problem:
   question: Is your problem one of these?
   answers:
-    - text: The boiler or pipes are making a loud noise
+    - text: I have no hot water - boiler in my home
 
-    - text: Electric heater is not working
-
-    - text: I have no hot water
+    - text: I have no hot water - communal boiler
 
     - text: Radiator is not working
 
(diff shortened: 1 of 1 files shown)


db/questions.yml
```diff
@@ -271,22 +269,27 @@ other_electrical_problem:
...
       desc: window_problem
 
 
 external_doors_problem:
-  question: Are you able to secure access to your property?
+  question: Is your problem one of these?
   answers:
-    - text: 'Yes'
-      desc: external_doors_problem
+    - text: Front or back door is stiff or loose - uPVC
```
leeky, line 285: Not sure if this would be better as two questions:

`Front or back door is stiff of loose` -> `What is your door made of?`

The second question could have `Wood` / `uPVC` and `I'm not sure` as answers, the latter directing them to the call centre for assistance(?)

JasonEwins: I'm being quite frank here with the identifying of material, let me test this with residents first rather than adding an additional question.

leeky (approved):
LGTM! :shipit:

PR #44, opened 2017-11-02
(183) Allow the Rails App to talk to the API

dgmstuart:
API requests are made with client certs.

This can't really be tested for real until the API access has been sorted out.

- Largely this is about allowing Faraday to be configured via environment variables
- Agreed with Leeky to skip testing that the client certs are actually used. This is unfortunate but we couldn't see how to do it, and if it doesn't work then we'll know very quickly since no requests will work.
- For completeness: I also considered applying the pattern of creating a 'Configuration' object which responds to methods, rather than being accessed like a Hash (which is the case with ENV). This would give us some protection against eg. typos in variable names, but is slightly more involved to work with (eg. in tests we'd need to pass in an object rather than a hash of params), so didn't seem worth it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,10 @@ The app should run successfully without these environment variables:
 
 #### Authentication
 
- - `HTTP_AUTH_USER` / `HTTP_AUTH_PASSWORD` - Set these to protect the site with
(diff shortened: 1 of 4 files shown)


leeky (approved):
LGTM! :shipit:

PR #46, opened 2017-11-03
(145) Send room in description

dgmstuart:
- Introduce 'mandatory' questions
- We talked about sending it as a special field, but this doesn't seem to have been locked in in the API standards, so send it in the description instead

### DIFF
--- a/app/models/question_set.rb
+++ b/app/models/question_set.rb
@@ -14,7 +14,8 @@ class QuestionSet
       raise UnknownQuestion, "Cannot find question with id '#{id}'"
     end
 
-    Question.new(id, @questions.fetch(id))
+    params = @questions.fetch(id).merge('id' => id)
+    Question.new(params)
   end
 
   class QuestionSetLoader
--- a/app/models/question_set.rb
+++ b/app/models/question_set.rb
@@ -4,19 +4,9 @@ class QuestionSet
   class BadlyFormattedYaml < Error; end
 
   def initialize(filename = 'db/questions.yml', partial_checker:)
-    yaml_data = File.read(Rails.root.join(filename))
-
-    begin
-      parsed_yaml = YAML.parse(yaml_data)
-    rescue => e
-      raise BadlyFormattedYaml, e.message
-    end
-
-    @questions = parsed_yaml.to_ruby
-
-    QuestionsValidator
-      .new(@questions, partial_checker: partial_checker)
-      .validate!
+    loader = QuestionSetLoader.new(partial_checker: partial_checker)
+    file_path = Rails.root.join(filename)
+    @questions = loader.load!(file_path)
   end
 
   def find(id)
@@ -25,5 +15,32 @@ class QuestionSet
     end
 
     Question.new(@questions.fetch(id))
+  end
+
+  class QuestionSetLoader
+    def initialize(partial_checker:)
+      @partial_checker = partial_checker
+    end
+
+    def load!(file_path)
+      yaml_data = File.read(file_path)
+      questions = parse_yaml(yaml_data).to_ruby
+
+      QuestionsValidator
+        .new(
+          questions,
+          partial_checker: @partial_checker,
+        ).validate!
+
+      questions
+    end
+
+    private
+
+    def parse_yaml(yaml_data)
+      YAML.parse(yaml_data)
+    rescue => e
+      raise BadlyFormattedYaml, e.message
+    end
   end
 end
--- a/app/models/question_set.rb
+++ b/app/models/question_set.rb
@@ -2,8 +2,6 @@ class QuestionSet
   class Error < StandardError; end
   class UnknownQuestion < Error; end
   class BadlyFormattedYaml < Error; end
-  class MissingQuestions < Error; end
-  class MissingPartials < Error; end
 
   def initialize(filename = 'db/questions.yml', partial_checker:)
     yaml_data = File.read(Rails.root.join(filename))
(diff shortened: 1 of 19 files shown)


app/models/question_set.rb
```diff
@@ -14,7 +14,7 @@ def find(id)
       raise UnknownQuestion, "Cannot find question with id '#{id}'"
     end
 
-    Question.new(@questions.fetch(id))
+    Question.new(id, @questions.fetch(id))
```
leeky: How about adding the `id` key to the `question_hash` when it gets loaded instead? (I'm not sure if I like how `id` need to be referenced twice in the initialiser)

```ruby
questions.each do |question, answers|
  [question, answers.merge!('id' => question)]
end
```

Also means you don't have to change any of the tests.

app/models/questions_validator.rb
```diff
@@ -30,6 +32,17 @@ def validate_desc
     raise MissingPartials, missing_partials if missing_partials.any?
   end
 
+  def validate_mandatory
```
leeky: :+1:

app/models/question_set.rb
```diff
@@ -26,4 +16,32 @@ def find(id)
 
     Question.new(@questions.fetch(id))
   end
+
+  class QuestionSetLoader
+    def initialize(partial_checker:)
+      @partial_checker = partial_checker
+      @root_url = Rails.root
```
leeky: `@root_url` doesn't appear to be used anywhere..? As an aside, `Rails.root` is the absolute path to the root of the app rather than a URL.

app/services/question_saver.rb
```diff
@@ -16,6 +16,19 @@ def save(form)
   private
 
   def persist_answers(form)
+    persist_room(form)
```
leeky, line 19: I'm wondering if these 'special' answers should be stored in a separate place?

`SelectedAnswerStore` is now storing: 1) the answers given to questions by a user *AND* 2) specific answers + sor code that are sent to the API when 'creating' a repair, which feel like different things to me.

app/models/repair_params.rb
```diff
@@ -4,9 +4,9 @@ def initialize(answers)
   end
 
   def problem
-    @answers.fetch('describe_repair').fetch('description').tap do |problem|
-      return 'n/a' if problem.blank?
-    end
+    return 'n/a' if description.blank?
```
leeky: Won't this cause the room information to be lost when description is blank..?

app/presenters/confirmation.rb
```diff
@@ -43,7 +43,7 @@ def scheduled_action
   end
 
   def room
-    @answers.fetch('room').fetch('room')
+    @answers.dig('room', 'room')
```
leeky: I'd forgotten about `dig`... :+1:

leeky (changes requested):

app/models/question_set.rb
dgmstuart: Oops! that was left over from the refactoring. 

Thanks for the aside! 👍 

Have amended this commit

app/models/repair_params.rb
dgmstuart: Good catch! I'll add a spec

app/presenters/confirmation.rb
dgmstuart: Me too!

app/services/question_saver.rb
dgmstuart: I agree, but I'm not sure what the improved solution would look like yet. 

I feel like what we've got now doesn't actually have any problems yet, so it's good enough for now, but I think we should keep an eye open for problems and refactoring to avoid them when we find them.

dgmstuart:
Rebased and addressed comments - @leeky: should be good for another look

spec/features/users_can_diagnose_their_issue_spec.rb
```diff
@@ -26,6 +26,7 @@
       .with('first')
       .and_return(
         Question.new(
+          'id' => 'fun',
```
leeky: So that the stub matches reality, shouldn't `id` be `first` in this case?

spec/features/users_can_diagnose_their_issue_spec.rb
```diff
@@ -43,6 +44,7 @@
       .with('second')
       .and_return(
         Question.new(
+          'id' => 'sure',
```
leeky: Same here - `id` should be `second`, to match the id that `#find` was sent.

spec/fixtures/basic_question.yml
```diff
@@ -1,4 +1,15 @@
 ---
+location:
```
leeky, line 2: Is it worth splitting this out into a new fixture? It now contains *three* questions, so is no longer just a 'basic question'.

spec/features/users_can_diagnose_their_issue_spec.rb
dgmstuart: Yup

spec/fixtures/basic_question.yml
dgmstuart: I could remove the final question, and then this is the smallest valid question?

leeky: That would be better :+1:

leeky: Ideally, it should be possible to overwrite the mandatory questions array in tests, so that the basic question is still valid. However, I'm not going to worry about this for now.

leeky (approved):

leeky:
@dgmstuart As I'd only suggested minor changes, I've gone ahead and added these myself, in order that I can merge is in your absence.

PR #47, opened 2017-11-10
(196) Enable manual downtime with feature flags

leeky:
Allow an administrator to disable the site when certain internal systems are unavailable. We'd prefer to direct users to the call centre, rather than fail halfway through the process with an error.

Eventually, we'll have more granular flags, that disable parts of the site, or provide alternative paths through the site (i.e. requesting a callback in the event that we are unable to book appointments).


leeky:
Rebased and updated with revised error page copy.

techbelly:
Initial comment (I still have to get the app running here) - I'm not sure that having a feature flag UI behind an http_auth password is a good solution here. Although, if it's already been talked about, ignore me.

On the one hand, it lets a service manager disable the service without developer intervention, which is a good thing. On the other, it's a password to be remembered and managed, reset when staff members leave, and so on. 

There's a danger it will end up on a post-it note somewhere and never be changed.

I do appreciate we don't have or need fully-fledged administrator user management and I don't think we should build it. I also get that Flipper comes with a baked-in UI to manage flags that reduces a bit of work.

What do you think about having a rake task to enable/disable the site instead? It avoids the whole user management problem - if you can execute a command on heroku, you can disable the site and hopefully deployment accounts will be better managed. But it does mean that it becomes a developer task to do this?

leeky:
@techbelly The problem with having a rake task, is that a developer at Hackney has to be added as a collaborator on this project and have the Heroku command line tool installed on their machine. There is a windows version of the Heroku CLI, however I'd have to check to see if someone needs to authorise the installation of software.

The suggestion is that they'll eventually be a "Hackney Service Status API" that you can ping to figure out what parts of the Hackney are down, and this will be used to take the site down.

I agree that protecting the UI behind basic auth is not ideal. I chose Flipper as we've used it in a couple of other projects, albeit both with persisted users.

techbelly:
@leeky Got it. 

There's no ideal right now, but lets make sure that we at least record these limitations in a backlog card proposing a better way of doing this kind of auth.

leeky:
@techbelly If I wasn't aware that we also want to feature flag the state where DRS is down but UH is up (allowing the user to report a repair, just without the appointment booking part), then I'd have likely just used Heroku's built-in maintenance mode to do this.

I'll add a card shortly.

techbelly:
Works well.

Can you just document the feature flag and the env variables required to protect the site in the project README, then I'll be happy to approve?

leeky:
@techbelly I've rebased and documented the feature flag in the README.

techbelly (approved):
All good.

PR #48, opened 2017-11-10
(176) Session is reset before reporting a new repair

leeky:
Reset the session when starting to work through the service a second
time. This ensures that users don't see previous answers in the forms
and also prevents a situation where the repair has both a `sor_code` and
a `callback_time`, which should be mutally exclusive.

NB: This users a controller spec rather that a feature spec, because the
contents of the encrypted session is not available from Capybara, so
there is no way of knowing if it was reset correctly.

### DIFF
--- a/app/controllers/questions/start_controller.rb
+++ b/app/controllers/questions/start_controller.rb
@@ -1,6 +1,8 @@
 module Questions
   class StartController < ApplicationController
     def index
+      reset_session
+
       @form = StartForm.new
     end
 
--- a/spec/features/users_can_report_a_repair_without_previous_answers_affecting_result_spec.rb
+++ b/spec/features/users_can_report_a_repair_without_previous_answers_affecting_result_spec.rb
@@ -90,6 +90,22 @@ RSpec.feature 'Users can report a repair without previous answers affecting the 
       )
     allow(QuestionSet).to receive(:new).and_return(fake_question_set)
 
+    # FIRST VISIT ========================
+    visit '/'
+    click_on 'Start'
+
+    # Emergency page:
+    choose_radio_button 'No'
+    click_on 'Continue'
+
+    # Fake decision tree
+    choose_radio_button 'Kitchen'
+    click_on 'Continue'
+
+    choose_radio_button 'Yes' # Gets a SOR code
+    click_on 'Continue'
+
+    # SECOND VISIT ========================
     visit '/'
     click_on 'Start'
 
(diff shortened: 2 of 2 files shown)


techbelly:
Initial comment (still getting the app running here) I think I would have liked to see a test here that focuses on the effect of reseting the session rather than, or perhaps in addition to, checking that reset session resets the session. 

I don't quite see the connection between resetting the session and 'prevents a situation where the repair has both a sor_code and a callback_time.' If those should be mutually exclusive, is there no way to enforce that through the choice of data structure we're using to represent the answers? Is it possible that this accidentally fixes something here that should be fixed (and thereby specified in the code) elsewhere?

leeky:
@techbelly I'll go back and add an integration test for this (although it's likely going to be very long indeed - it needs to go through the entire flow twice to verify the situation).

As you go through the repair reporting process, certain answers are stored away in the SelectedAnswerStore (basically the session cookie) for later use. At one point we branch based on whether we've stored a SOR code - i.e. we show either the time the user has requested a callback OR the time they booked an appointment for.

If the session doesn't get reset, the user may have a SOR code stored from the previous run through the repair site in addition to a callback_time from the current run through. Since I started writing this response, I've realised that we when setting a callback_time we could wipe any previously stored SOR code - but I think there's additional stories that bring about the possibility of these attributes existing together for valid reasons.

I agree that there's probably work that could be done to tidy the data structures, but I'm not sure we have time to do that. It's not perfect, but fixes the problem quickly.

techbelly:
@leeky Understood. 

I think a test that verifies the behaviour would be good, is there no simple way to pass in a smaller question tree? That would seem a useful thing to be able to do.

In any case, a test like this would have value now and will help us when/if we refactor the data structures later.

leeky:
@techbelly I think the use case we're trying to prevent is when the user previously chose a callback time, and the next time around picks an appointment. However, this requires https://github.com/LBHackney-IT/maintain-my-home/pull/42 to be merged in order to make any sense. Can you take a look at that one?

dgmstuart:
@leeky #42 is now merged - what needs to happen next?

leeky:
@dgmstuart I need to write a feature spec that clicks through the form to a callback outcome, and then again with a diagnosis outcome, checking that the confirmation page is in the correct state - i.e. shows "appointment is booked" rather than "we will call you".

app/controllers/questions/start_controller.rb
```diff
@@ -1,6 +1,8 @@
 module Questions
   class StartController < ApplicationController
     def index
+      reset_session
```
dgmstuart, line 4: Do we want to reset the whole session, or just the selected answers?

spec/controllers/questions/start_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Questions::StartController do
```
dgmstuart: Now that we have a feature spec, do we need this spec? What would happen if we switched over to using a different store for the selected answers?

spec/features/users_can_report_a_repair_without_previous_answers_affecting_result_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+
+    # Emergency page:
+    choose_radio_button 'No'
+    click_on 'Continue'
+
+    # Fake decision tree
+    choose_radio_button 'Kitchen'
+    click_on 'Continue'
+
+    choose_radio_button 'No' # Doesn't get a SOR code
```
dgmstuart: Currently this spec verifies the behaviour by checking what is shown on the confirmation page. Would it be clearer and shorter to check that when the user gets an SOR code _first_ then they see the appointments page when requesting a second repair that doesn't have an SOR code?

dgmstuart (changes requested):
It would be good to dig into the approach here a bit more.

app/controllers/questions/start_controller.rb
lexijax: Yes, should be the whole session - no defined need for anything more granular at this point.

spec/controllers/questions/start_controller_spec.rb
leeky: :+1:

leeky:
@dgmstuart I've removed the original controller spec, as suggested. Also, I've added an additional commit that tidies up the feature spec, so that it does the least amount of work possible to verify that the session is getting reset correctly.

app/controllers/questions/start_controller.rb
dgmstuart: @lexijax My thought was that I'm not sure what else gets stored in the session (the principle being don't mess with what you don't control) and that this might have implications for analytics or rate limiting. I don't know that though.

leeky: @dgmstuart `reset_session` only deletes the cookie that Rails uses. Analytics etc will use their own separate cookies.

dgmstuart: @leeky Indeed. My thought was that maybe hotjar uses it to identify unique users. Probably not

dgmstuart (approved):

PR #49, opened 2017-11-10
(199) Assign signed-off SOR codes to outcomes

leeky:
Add all the SOR codes that exist on the spreadsheet and have been signed
off by the relevant people.


lexijax (approved):
Checked a fair few of these against the spreadsheet and they look fine.

PR #51, opened 2017-11-15
Update lighting & hot water description

JasonEwins:
1. Updated lighting flow to pose question prior to replacing light bulb message
2. Updated 'Heating' options to 'Heating and Hot water' as per usability findings

### DIFF
--- a/db/questions.yml
+++ b/db/questions.yml
@@ -93,7 +93,7 @@ kitchen_problem:
     - text: Electrical
       next: kitchen_electrical_problem
 
-    - text: Heating
+    - text: Heating or Hot water
       next: heating_problem
 
     - text: Sink
@@ -190,7 +190,7 @@ bathroom_problem:
     - text: Electrical
       next: bathroom_electrical_problem
 
-    - text: Heating
+    - text: Heating or Hot water
       next: heating_problem
 
     - text: Toilet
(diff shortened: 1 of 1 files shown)


db/questions.yml
```diff
@@ -93,7 +93,7 @@ kitchen_problem:
     - text: Electrical
       next: kitchen_electrical_problem
 
-    - text: Heating
+    - text: Heating or Hot water
```
leeky, line 96: How about if we split these into two options, so it's easier to scan down the page?

db/questions.yml
```diff
@@ -376,11 +376,21 @@ light_check_trip_switches:
...
+      next: light_check_light_bulb
 
     - text: 'No'
       page: check_fusebox
 
+light_check_light_bulb:
+  question: Have you replaced the light bulb?
+  answers:
+    - text: 'Yes'
+      desc: lightfitting_problem
```
leeky, line 388: What custom text should appear on this page?

db/questions.yml
JasonEwins: That makes perfect sense, happy for that. 

If we can include the 'I have no heating' option within the 'heating' options as well, incase user doesn't pick up on initial emergency filter page

db/questions.yml
JasonEwins: Content for page as per copy document - row 38 column E (with Louise's revisions)

leeky:
Closing this in favour of a new branch which will also contain Louise's copy changes.

PR #52, opened 2017-11-17
(136) Hide service from web crawlers

leeky:
We want users of the service to find us via the main hackney.gov.uk web site, so block web crawlers from indexing the site.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,1 +1,3 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+# Users should only find the service through the main Hackney web site
+User-agent: *
+Disallow: /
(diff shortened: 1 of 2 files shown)


public/robots.txt
```diff
@@ -1 +1,3 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+# Users should only find the service through the main Hackney web site
```
dgmstuart, line 1: Good to have this comment 👍

dgmstuart (changes requested):
The trello card said that submit buttons should have `rel='nofollow'` - should we do that too?

leeky:
@dgmstuart The `rel` attribute is only available on `<a>`, `<area>` and `<link>` tags. Google also doesn't index pages that are accessible via a POST (such a form submissions).

dgmstuart:
@leeky Ah right - makes sense. Is it worth adding that comment to the Trello card for future reference (or even the commit?)

dgmstuart (approved):

leeky:
I've rebased and updated the relevant commit with a note about buttons and nofollow.

PR #53, opened 2017-11-20
(202) Properties endpoint fixes

leeky:
As the Hackney API will shortly be brought inline with the [agreed API standards](https://github.com/LBHackney-IT/api-standards) we need to update our client code which consumes them.

This PR updates the `GET /v1/properties?postcode=` and `GET /v1/properties/...` so they provide the correct parameters in the request and parse the responses correctly.

- Rename `short_address` to `address`
- Rename `property_reference` to `propertyReference`
- Add missing `postcode` to several stubs
- Wrap multiple results in a `results` key
- Prefix endpoints with `/v1`

### DIFF
--- a/app/queries/hackney_api.rb
+++ b/app/queries/hackney_api.rb
@@ -4,11 +4,11 @@ class HackneyApi
   end
 
   def list_properties(postcode:)
-    @json_api.get('properties?postcode=' + postcode)
+    @json_api.get('v1/properties?postcode=' + postcode)
   end
 
   def get_property(property_reference:)
-    @json_api.get('properties/' + property_reference)
+    @json_api.get('v1/properties/' + property_reference)
   end
 
   def create_repair(repair_params)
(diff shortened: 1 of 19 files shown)


app/queries/hackney_api.rb
```diff
@@ -1,10 +1,16 @@
...
     @json_api = json_api
   end
 
   def list_properties(postcode:)
-    @json_api.get('v1/properties?postcode=' + postcode)
+    response = @json_api.get('v1/properties?postcode=' + postcode)
+
+    raise MissingResults unless response.is_a?(Hash) && response.key?('results')
+
+    response['results']
```
dgmstuart: Did you consider using `fetch`? That raises a `KeyError` if the key is not found.

dgmstuart (approved):
Good job - sorry this required so many changes - I wonder if we could have designed things differently in the first place to make this change smaller?

app/queries/hackney_api.rb
leeky: :+1:

PR #54, opened 2017-11-22
(214) Include callback time in description

dgmstuart:
Format it in the same way as it is displayed on the confirmation page.

Includes fix: Callback time gets saved as a symbol, and read out as a
symbol (we're expecting a string key). This is because it hasn't been saved
to the session cookie and read out again (that's what casts the symbols to
strings).

We should probably write everything as strings, since the casting of
symbols to strings is a little confusing, but that should happen separately
to this PR.

### DIFF
--- a/app/models/repair_params.rb
+++ b/app/models/repair_params.rb
@@ -4,14 +4,11 @@ class RepairParams
   end
 
   def problem
-    if description.present? && room.present?
-      "#{description} (Room: #{room})"
-    elsif description.present?
-      description
-    elsif room.present?
-      "Room: #{room}"
-    else
-      'n/a'
+    lines = []
+    lines << description if description.present?
+    lines << "Room: #{room}" if room.present?
+    lines.compact.join("\n\n").tap do |output|
+      return 'n/a' if output.empty?
     end
   end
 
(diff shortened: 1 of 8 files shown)


app/models/repair_params.rb
```diff
@@ -4,15 +4,10 @@ def initialize(answers)
...
-    if description.present? && room.present?
-      "#{description} (Room: #{room})"
-    elsif description.present?
-      description
-    elsif room.present?
-      "Room: #{room}"
-    else
-      'n/a'
-    end
+    lines = [description]
```
leeky, line 7: Much more readable! :tada:

app/models/repair_params.rb
```diff
@@ -34,10 +29,19 @@ def diagnosed?
   private
 
   def description
-    @answers.fetch('describe_repair').fetch('description')
+    @answers.fetch('describe_repair').fetch('description').tap do |desc|
```
leeky, line 32: I really must learn to reach for `tap` myself :+1:

app/presenters/callback.rb
```diff
@@ -1,27 +1,37 @@
...
-      raise InvalidCallbackTimeError
-    end
+    @time_slot.to_s
   end
 
   def to_partial_path
     '/confirmations/callback'
   end
+
+  class TimeSlot
```
leeky, line 17: Nice abstraction :+1:

leeky (approved):
LGTM! :shipit:

PR #55, opened 2017-11-23
(216) Repairs endpoint fixes

leeky:
Update the `GET /v1/repairs/:repairRequestReference` and `POST /v1/repairs/` so they provide the correct parameters in the request and parse the responses correctly.

- Rename `requestReference` to `repairRequestReference`
- Rename `problem` to `problemDescription`
- Change from `repairOrders` with a `jobCode` to `workOrders` with a `sorCode`
- Rename `orderReference` to `workOrderReference`, and extract this differently as it's now nested inside the `workOrders` array in the response
- Include contact `name `and `telephoneNumber` when submitting a repair (NB: callback time is being added separately in #54)
- Prefix the repairs routes with `v1/` as per the agreed standards

### DIFF
--- a/app/models/repair.rb
+++ b/app/models/repair.rb
@@ -4,9 +4,7 @@ class Repair
   end
 
   def request_reference
-    @repair_data.fetch('requestReference') do |_key|
-      @repair_data.fetch('repairRequestReference')
-    end
+    @repair_data.fetch('repairRequestReference')
   end
 
   def work_order_reference
--- a/app/models/repair.rb
+++ b/app/models/repair.rb
@@ -8,6 +8,8 @@ class Repair
   end
 
   def work_order_reference
-    @repair_data['orderReference']
+    return nil unless @repair_data.key?('workOrders')
+
+    @repair_data.fetch('workOrders')&.first&.fetch('workOrderReference')
   end
 end
--- a/app/models/repair_params.rb
+++ b/app/models/repair_params.rb
@@ -3,7 +3,7 @@ class RepairParams
     @answers = answers
   end
 
-  def problem
+  def problem_description
     lines = [description]
     lines << "Room: #{room}" if room.present?
     lines << "Callback requested: between #{callback_time}" if callback_time
--- a/app/models/repair_params.rb
+++ b/app/models/repair_params.rb
@@ -26,6 +26,14 @@ class RepairParams
     sor_code.present?
   end
 
+  def contact_full_name
+    @answers.fetch('contact_details').fetch('full_name')
+  end
+
+  def contact_telephone_number
+    @answers.fetch('contact_details').fetch('telephone_number')
+  end
+
   private
 
   def description
--- a/app/queries/hackney_api.rb
+++ b/app/queries/hackney_api.rb
@@ -13,11 +13,11 @@ class HackneyApi
   end
 
   def create_repair(repair_params)
-    @json_api.post('repairs', repair_params)
+    @json_api.post('v1/repairs', repair_params)
   end
 
   def get_repair(repair_request_reference:)
-    @json_api.get('repairs/' + repair_request_reference)
+    @json_api.get('v1/repairs/' + repair_request_reference)
   end
 
   def list_available_appointments(work_order_reference:)
--- a/app/services/contact_details_saver.rb
+++ b/app/services/contact_details_saver.rb
@@ -16,9 +16,9 @@ class ContactDetailsSaver
 
   def persist_answers(form)
     @selected_answer_store.store_selected_answers(
-      :contact_details,
-      full_name: form.full_name,
-      telephone_number: form.telephone_number,
+      'contact_details',
+      'full_name' => form.full_name,
+      'telephone_number' => form.telephone_number
     )
   end
 end
--- a/spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
+++ b/spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
@@ -11,7 +11,7 @@ RSpec.feature 'Resident can see a confirmation of their repair request' do
     allow(fake_api).to receive(:get).with('v1/properties?postcode=E5 8TE').and_return('results' => [property])
     allow(fake_api).to receive(:get).with('v1/properties/00000503').and_return(property)
     allow(fake_api).to receive(:post)
-      .with('repairs', anything)
+      .with('v1/repairs', anything)
       .and_return(
         'repairRequestReference' => '00367923',
         'priority' => 'N',
@@ -25,7 +25,7 @@ RSpec.feature 'Resident can see a confirmation of their repair request' do
         ]
       )
     allow(fake_api).to receive(:get)
-      .with('repairs/00367923')
+      .with('v1/repairs/00367923')
       .and_return(
         'repairRequestReference' => '00367923',
         'priority' => 'N',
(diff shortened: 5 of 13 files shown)


app/models/repair.rb
```diff
@@ -4,9 +4,7 @@ def initialize(repair_data)
   end
 
   def request_reference
-    @repair_data.fetch('requestReference') do |_key|
```
dgmstuart, line 7: There's a spec for the second case here which needs deleting (or at least amending) - the context is `'when there is a reference with an outdated key'`

app/models/repair_params.rb
```diff
@@ -26,6 +26,14 @@ def diagnosed?
     sor_code.present?
   end
 
+  def contact_full_name
+    @answers.dig('contact_details', 'full_name')
+  end
+
+  def contact_telephone_number
+    @answers.dig('contact_details', 'telephone_number')
```
dgmstuart: Dig works well for optional data, because if keys are missing then it returns nil. 
Is that what we want here, or do we want to raise an exception if contact details are missing, like we do with `propertyReference`? Do we expect all repairs have this data mandatory?

app/services/contact_details_saver.rb
```diff
@@ -16,9 +16,9 @@ def save(form)
 
   def persist_answers(form)
     @selected_answer_store.store_selected_answers(
-      :contact_details,
-      full_name: form.full_name,
-      telephone_number: form.telephone_number,
+      'contact_details',
+      'full_name' => form.full_name,
+      'telephone_number' => form.telephone_number
```
dgmstuart, line 21: Would this benefit from a comment in the commit message to explain it?

spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
```diff
@@ -273,6 +277,10 @@
       priority: 'N',
       problemDescription: "My sink is blocked\n\nRoom: Other\n\nCallback requested: between 8am and 5pm",
       propertyReference: '00000503',
+      contact: {
+        name: 'John Evans',
+        telephoneNumber: '078 98765 432',
+      },
```
dgmstuart, line 293: Does this data come back from the `get_repair` requests? If so should we update our stubs (above) to reflect this?

spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
```diff
@@ -149,7 +149,7 @@
     expect(fake_api).to have_received(:post).with(
       'repairs',
       priority: 'N',
-      problem: "My sink is blocked\n\nRoom: Kitchen",
+      problemDescription: "My sink is blocked\n\nRoom: Kitchen",
```
dgmstuart: Are we also getting this new key back in the responses? If so should we update our stubs (above) so that they are more realistic? Currently they include the `problem` key.

spec/presenters/confirmation_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'active_support/inflector'
```
dgmstuart, line 2: Is this required?

app/models/repair.rb
```diff
@@ -8,6 +8,8 @@ def request_reference
   end
 
   def work_order_reference
-    @repair_data['orderReference']
+    return nil unless @repair_data.key?('workOrders')
+
+    @repair_data.fetch('workOrders')&.first&.fetch('workOrderReference')
```
dgmstuart, line 13: Could also do 

```ruby
@repair_data.fetch('workOrders', [{}]).first.fetch('workOrderReference')
```

...but it's maybe actually less readable.

dgmstuart: On a different note: we said in the [api standards](https://github.com/LBHackney-IT/api-standards#response-2) that:

> If the repair was created without work orders, an empty array of workOrders should be returned (rather than no key or a null value)

...so maybe it _should_ blow up if the `'workOrders'` isn't an Array? Really not sure either way.

spec/models/repair_spec.rb
```diff
@@ -25,6 +32,20 @@
         expect(Repair.new(repair_data).work_order_reference).to be_nil
       end
     end
+
+    context 'when workOrders is returned, but is empty' do
```
dgmstuart: Good test case 👍

app/models/repair.rb
```diff
@@ -8,6 +8,8 @@ def request_reference
   end
 
   def work_order_reference
-    @repair_data['orderReference']
```
dgmstuart, line 13: Great commit message 👍

app/queries/hackney_api.rb
```diff
@@ -13,11 +13,11 @@ def get_property(property_reference:)
   end
 
   def create_repair(repair_params)
-    @json_api.post('repairs', repair_params)
+    @json_api.post('v1/repairs', repair_params)
```
dgmstuart, line 16: We can't do it now, but do you think we should ultimately make the version part of the API root?

dgmstuart (changes requested):
Nice work 👍 

My main concern is about making sure our stubs in the big feature specs are realistic. It's a bit tricky since we're not using most of the data returned from create and get repair, so nothing is failing and forcing us to update these.

app/models/repair.rb
leeky: Good spot! I'll remove that old test.

app/queries/hackney_api.rb
leeky: I completely agree :+1: 

Thinking about this, I was imagining scenarios where you'd want to use `v2` of a particular endpoint to get additional data back, but `v1` of everything else - but in reality, new versions  should maintain compatibility as much as possible.

app/services/contact_details_saver.rb
leeky: Yes, I've added a comment about this. Also added a card around fixing `SelectedAnswerStore` so that stringified/symbolized keys behave the same way.

spec/features/residents_can_see_confirmation_of_repair_request_spec.rb
leeky: Possibly. We're not currently using the retrieved data (from `get_repair`) on the confirmation page, but maybe should do. The data in the session and the data we get back from UH should be identical anyway.

Maybe one for another card which uses more of the data returned from the API rather than parroting back the user input(?)

spec/presenters/confirmation_spec.rb
leeky: Because this spec doesn't include `rails_helper`, if you run these specs in isolation you get an error because `ActiveSupport::Inflector` isn't loaded.

I've added this note into the commit message.

app/models/repair_params.rb
leeky: :+1:

app/models/repair.rb
leeky: Yeah, I tried putting it into one line, but it felt less readable.

Whilst the API standards says this, there's no practical difference between an empty `workOrders` array and the lack of such an array.

I'd prefer to err on the side of not blowing the site up and annoying the users in this case.

leeky:
@dgmstuart Should be good for another look.

spec/presenters/confirmation_spec.rb
dgmstuart: Ah I see: this is because `Confirmation` directly references `AppointmentsPresenter` which uses `to_sentence`. I guess this is a sign that our test of Confirmation is actually integration testing the presenter classes as well. I don't think there's any real benefit to addressing that now though.

dgmstuart (approved):

PR #56, opened 2017-11-23
(201) Use 'midday' instead of '12pm'

dgmstuart:
As per the gov.uk style guide:
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times

Also use 'to' instead of dashes in time ranges

### DIFF
--- a/app/presenters/appointment_presenter.rb
+++ b/app/presenters/appointment_presenter.rb
@@ -13,7 +13,7 @@ class AppointmentPresenter
   end
 
   def description
-    "#{day_of_week} #{begin_time}-#{end_time} (#{day} #{month})"
+    "#{day_of_week} #{begin_time} to #{end_time} (#{day} #{month})"
   end
 
   def date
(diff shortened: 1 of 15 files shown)


dgmstuart:
rebased

leeky:
@dgmstuart Should we do the same when listing available appointments? Given the ongoing fixes to the API endpoints this feels like a separate card perhaps..?

app/presenters/appointment_presenter.rb
```diff
@@ -59,6 +59,7 @@ def end_time
   end
 
   def friendly_time(time)
+    return 'midday' if time == time.midday
```
leeky, line 62: Rails has *almost* too many convenience methods :+1:

leeky (approved):
LGTM!

My only comment was whether we should apply these changes in appointment booking and static pages here or create new cards..?

dgmstuart:
@leeky Can this be merged? Were there some places I missed?

leeky:
@dgmstuart There are a few places static pages which have time ranges separated by `&ndash;` instead of ` to `. I'll add these, rebase and merge this in :+1:

PR #57, opened 2017-11-27
(230) Rails app fixes for appointments endpoints

leeky:
**NB: This is based off #216, which should be merged first**

  - Fetching a list of appointments is now done via `GET /v1/work_orders/:workOrderReference/available_appointments`. The response is wrapped in a `results` attribute.
  - After booking an appointment, the response no longer contains the `status` attribute, because this doesn't actually get returned by DRS.


dgmstuart (approved):
LGTM 🚢

PR #58, opened 2017-11-27
(235) Rationalise all phone numbers

dgmstuart:
There are a lot of pages in the site which include phone numbers. In some
 cases the numbers were inconsistent, and how they were presented on the
 page was inconsistent as well.

 Now there is only one place where phone numbers are explicitly written, and
 everywhere they are used in the pages they are used in the same way: as a
 telephone number link with the appropriate 'itemprop'.

 Note that in some places the specific number used on a page has been
 corrected to the most appropriate one.

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -1,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationHelper do
+  describe '#telephone_number' do
+    it 'wraps the phone number in appropriate markup' do
+      expect(helper.telephone_number('01234 567 890'))
+        .to eq(%(<span itemprop="telephone"><a href="tel:01234567890">01234567890</a></span>))
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


dgmstuart:
rebased

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationHelper do
+  describe '#telephone_number' do
+    it 'wraps the phone number in appropriate markup' do
+      expect(helper.telephone_number('01234 567 890'))
+        .to eq(%(<span itemprop="telephone"><a href="tel:01234567890">01234567890</a></span>))
```
leeky: Shouldn't this test fail? The code is `ApplicationHelper` only removed the spaces from the `href` and not the string contained within - so `helper.telephone_number('01234 567 890')` should generate:

```html
... <a href="tel:01234567890">01234 567 890</a> ...
```

leeky (changes requested):

spec/helpers/application_helper_spec.rb
dgmstuart: Good point! Will investigate

dgmstuart: Ok, that was weird: I was (incorrectly) using `gsub!` instead of `gsub`, and I guess that gets evaluated first because in the markup href comes before the text.

Damn mutation!

dgmstuart:
@leeky Rebased and pushed a commit to fix

leeky (approved):

leeky:
LGTM! :shipit:

PR #59, opened 2017-11-27
(208) User facing validation

leeky:
Add validation in the following places, to ensure that user input won't cause an error from the API.

 - Contact name is at least 2 characters
 - Telephone number is 10-11 digits (spaces and non-numeric values don't count towards this)
 - Limit description to 500 characters, and show "characters remaining" using JavaScript

character_count

### DIFF
--- a/app/assets/javascripts/character_limit.js
+++ b/app/assets/javascripts/character_limit.js
@@ -1,0 +1,37 @@
+$(document).ready(function(){
+  $('textarea').each(function() {
+    var textArea = $(this);
+
+    var limitElement = $("<div>",
+      {
+        "class" : "character-limit",
+        "aria-live" : "polite"
+      }
+    );
+
+    textArea.parent().append(limitElement);
+
+    function updateRemaining() {
+      var limit = textArea.attr('maxlength');
+      var used = textArea.val().length;
+
+      var remaining = limit - used;
+
+      if (remaining == 1) {
+        remainingText = "1 character remaining";
+      } else {
+        remainingText = remaining + " characters remaining";
+      }
+
+      limitElement.text(remainingText);
+    }
+
+    if (textArea.attr('maxlength')) {
+      textArea.keyup(function() {
+        updateRemaining();
+      });
+
+      updateRemaining();
+    }
+  });
+});
--- a/app/models/contact_details_form.rb
+++ b/app/models/contact_details_form.rb
@@ -6,8 +6,24 @@ class ContactDetailsForm
   validates :full_name, presence: true, length: { minimum: 2 }
   validates :telephone_number, presence: true
 
+  validate :telephone_number_format
+
   def initialize(hash = {})
     @full_name = hash[:full_name]
     @telephone_number = hash[:telephone_number]
+  end
+
+  private
+
+  def telephone_number_format
+    return unless telephone_number
+
+    telephone_number_only_digits = telephone_number.gsub(/\D/, '')
+
+    if telephone_number_only_digits.size < 10
+      errors.add(:telephone_number, :too_short, count: 10)
+    elsif telephone_number_only_digits.size > 11
+      errors.add(:telephone_number, :too_long, count: 11)
+    end
   end
 end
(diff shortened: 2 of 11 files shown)


app/models/contact_details_form.rb
```diff
@@ -3,7 +3,8 @@ class ContactDetailsForm
 
   attr_reader :full_name, :telephone_number
 
-  validates :full_name, :telephone_number, presence: true
+  validates :full_name, presence: true, length: { minimum: 2 }
```
dgmstuart, line 6: Do we also need a maximum on this? It can be large (eg. 100?), but could things go wrong if we don't specify a limit?

app/models/contact_details_form.rb
```diff
@@ -6,8 +6,24 @@ class ContactDetailsForm
...
+  def telephone_number_format
+    return unless telephone_number
+
+    telephone_number_only_digits = telephone_number.gsub(/\D/, '')
+
+    if telephone_number_only_digits.size < 10
+      errors.add(:telephone_number, :too_short, count: 10)
+    elsif telephone_number_only_digits.size > 11
+      errors.add(:telephone_number, :too_long, count: 11)
+    end
```
dgmstuart, line 27: Do we need a more forgiving length range?

The following would fail this:
- 0044 7812 345 678
- +44 7812 345 679
- +44 (0) 7812 345 679

Looks like 10 is the smallest valid UK personal phone number, but numbers in some other countries could be shorter (7 including the area code?)- though it's unlikely that we would come across those.

How about 7-20?

spec/models/describe_repair_form_spec.rb
```diff
@@ -13,5 +13,13 @@
       form = DescribeRepairForm.new(description: 'my taps are leaking')
       expect(form).to be_valid
     end
+
+    it 'is invalid with an overly long description' do
+      form = DescribeRepairForm.new(description: ('X' * 501))
```
dgmstuart, line 18: Nice 👍

spec/models/describe_unknown_repair_form.rb
```diff
@@ -2,11 +2,24 @@
...
+
+    it 'is invalid without a description' do
       form = DescribeUnknownRepairForm.new
       form.valid?
 
       expect(form.errors.details[:description]).to include(error: :blank)
     end
+
+    it 'is invalid with an overly long description' do
+      form = DescribeUnknownRepairForm.new(description: ('X' * 201))
```
dgmstuart, line 19: 201? How come this passes?

app/assets/javascripts/character_limit.js
```diff
@@ -0,0 +1,37 @@
+$(document).ready(function(){
+  $('textarea').each(function() {
+    var textArea = $(this);
+
+    var limitElement = $("<div>",
+      {
+        "class" : "character-limit",
+        "aria-live" : "polite"
```
dgmstuart, line 8: Mad skillz 🤘

dgmstuart (changes requested):
The Javascript updating is really nice. 

I wonder if the telephone length restrictions are too tight? What are we defending against?

spec/models/describe_unknown_repair_form.rb
leeky: Good point... I will investigate :+1:

leeky: @dgmstuart I figured out why this was passing... it wasn't.

The spec file is `spec/models/describe_unknown_repair_form.rb`, which is missing the `_spec` suffix, so doesn't get run. If I run the spec on it's own it fails as expected!

leeky: I've rebased and renamed this file to `spec/models/describe_unknown_repair_form_spec.rb` and immediately saw a failing spec that I've now fixed.

techbelly:
Good spot, this is why it's always good to start with a failing test!

leeky:
@techbelly Annoyingly, this is what I did originally. However I didn't think to make a failing spec when I went back and changed the upper limit from 200 to 500 characters after the fact! I just made the change, and ran the whole suite, which passed, so assumed all was well!

dgmstuart (approved):

dgmstuart:
LGTM 🚢

dgmstuart:
rebased

PR #60, opened 2017-11-27
(172) Display friendly messages when api fails

dgmstuart:
- Raise errors for various kinds of failure
- Display a page to the user
- Ensure that these are logged

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,4 +22,8 @@ class ApplicationController < ActionController::Base
   def check_service_status
     raise ServiceDisabled if App.flipper.enabled?(:service_disabled)
   end
+
+  rescue_from JsonApi::ApiError do
+    render 'errors/api_error'
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,11 @@ class ApplicationController < ActionController::Base
     render 'errors/service_disabled'
   end
 
+  rescue_from JsonApi::ApiError do |e|
+    logger.error "[Handled] #{e.class}: #{e.message}"
+    render 'errors/api_error'
+  end
+
   include Authentication
 
   def selected_answer_store
@@ -21,9 +26,5 @@ class ApplicationController < ActionController::Base
 
   def check_service_status
     raise ServiceDisabled if App.flipper.enabled?(:service_disabled)
-  end
-
-  rescue_from JsonApi::ApiError do
-    render 'errors/api_error'
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -9,6 +9,11 @@ class ServiceDisabled < StandardError; end
     render 'errors/service_disabled'
   end
 
+  rescue_from JsonApi::ApiError do |e|
+    logger.error "[Handled] #{e.class}: #{e.message}"
```
leeky, line 13: Really like the `[Handled]` prefix here :+1:

spec/queries/json_api_spec.rb
```diff
@@ -169,6 +169,47 @@
...
     end
 
+    context 'when the response error json is in the wrong format' do
+      context 'when the response is a 400 status' do
+        it 'raises an error' do
+          json_api = JsonApi.new(api_root: 'http://hackney.api:8000')
+          stub_request(:get, 'http://hackney.api:8000/properties?postcode=A1%201AA')
+            .to_return(
+              status: 400,
+              body: [{ 'userMessage' => 'That was invalid', 'developerMessage' => 'invalid request' }].to_json
```
leeky, line 179: Is it worth having one of these return multiple items in the array - or did we conclude this would never happen..?

app/queries/json_api.rb
```diff
@@ -1,27 +1,73 @@
...
+  end
+
+  private
+
+  def call_and_raise
+    yield.tap { |response| raise_for_http_status(response) }
   rescue Faraday::ParsingError => e
     raise InvalidResponseError, e.message
+  rescue Faraday::ConnectionFailed => e
+    raise ConnectionError, e
```
leeky: Why is this exception re-raised with all the exception metadata, whereas the above exception "InvalidResponseError" is re-raised passing only `e.message`.

leeky (approved):
This all looks good. I've just made a couple of comments.

app/queries/json_api.rb
dgmstuart: Good spot - this should be made consistent.

It actually has the same result: the second argument to raise is a string, so it casts the error to a string, which returns the same value as e.message.

dgmstuart:
rebased

app/controllers/application_controller.rb
dgmstuart: Thanks :) 

Just for interest: there's actually support for 'tagged' errors, but this didn't seem like the intended use case for them: http://api.rubyonrails.org/classes/ActiveSupport/TaggedLogging.html#method-i-tagged

app/queries/json_api.rb
dgmstuart: Amended the commit

spec/queries/json_api_spec.rb
dgmstuart: Yeah I think it won't, but could in theory. I think it's worth a spec: i'll add one in that commit

PR #61, opened 2017-11-28
(210) Prevent appointment tampering

leeky:
Because appointments returned from the API don't have a simple id to reference them, I used the `beginDate` and `endDate` concatenated together instead.

This comes with a major drawback - we're trusting the contents of these fields between coming out of the API, being used in the user's browser and being used to book an appointment. So there's a chance that the user could manipulate the time range to their advantage.

Using `ActiveSupport::MessageVerifier` I encrypt the appointment hash and use that as the 'value' for the radio button in the appointment form. This data is then decrypted and then sent to the API in order to book the appointment.

Now, if someone attempts to alter the radio button value, an exception is thrown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ root of the application containing environment variable assignments.
 The following environment variables are required to run the site:
 
 - `HACKNEY_API_ROOT` - the root of the Hackney API which is used by the site
+- `ENCRYPTION_SECRET` - secret used to prevent parameter tampering
 
 ### Optional config
 
--- a/app/models/appointment_form.rb
+++ b/app/models/appointment_form.rb
@@ -1,43 +1,28 @@
 class AppointmentForm
-  class InvalidAppointment < StandardError; end
-
   include ActiveModel::Model
 
   attr_reader :appointment_id
 
   validates :appointment_id, presence: true
 
-  def initialize(hash = {})
-    @appointment_id = hash[:appointment_id]
-
-    validate_appointment_dates!
+  def initialize(hash = {}, verifier: ActiveSupport::MessageVerifier)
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -12,6 +12,7 @@ root of the application containing environment variable assignments.
 The following environment variables are required to run the site:
 
 - `HACKNEY_API_ROOT` - the root of the Hackney API which is used by the site
+- `ENCRYPTION_SECRET` - a shared secret used to prevent parameter tampering
```
dgmstuart: "shared secret" usually refers to something shared between two different parties/applications/servers - I read this assuming it was a secret shared with the API, but looking at the code that isn't the case?

app/models/appointment_form.rb
```diff
@@ -1,43 +1,28 @@
...
-
-    split_date.each do |date|
-      begin
-        Time.zone.parse(date)
-      rescue ArgumentError
-        raise InvalidAppointment
-      end
-    end
+  def decrypted_appointment
+    @decrypted_appointment ||= @verifier
```
dgmstuart, line 24: Is there some duplication here around how we initialise the verifier? What would the tests look like if we initialised it in the controller and injected an instance?

dgmstuart (approved):
Looks good - I think there might be an opportunity to simplify things around the verifier - up to you.

Great commit message 👍

leeky:
@dgmstuart FYI I'm going to merge this as-is, as I ran out of time refactoring to remove duplication. I've push up a WIP branch, and will add a card in case we want to revisit it in the future.

README.md
leeky: :+1:

PR #62, opened 2017-11-29
(242) Handle error responses without root key

dgmstuart:
This is expected to be reverted later

Errors are supposed to have a root key of 'errors'. Currently they
don't, but we still want to be able to handle them.


dgmstuart:
Eek -the branch name is wrong. will fix and re-raise

PR #63, opened 2017-11-30
(255) Check that all referenced pages exist

dgmstuart:
We found in testing that the 'blocked_only_toilet' page doesn't exist, so
we need a check for pages to make sure we don't have any others.

This is expected to fail since there's one test which actually loads the real questions and there is one missing page. Should be reviewable now and mergeable when that page is created


leeky:
Rebasing.

leeky (approved):
LGTM - the tests were failing up until the point where the missing page was added back, as expected :shipit:

PR #66, opened 2017-11-30
(251) Stop mobile Safari adding links to reference numbers

leeky:
Add a meta tag into the `<head>` of all pages to prevent Safari on iOS from auto-detecting phone numbers. We're already adding `tel:...` links to all phone numbers, and this was affecting the output of repair reference numbers.

fixed-ios-reference-number

Tested in iPhone SE Safari through BrowserStack.


dgmstuart (approved):
Well done - LGTM 👍

PR #68, opened 2017-12-01
(247) Users should call RCC if they cannot make any appointment slot

leeky:
If the user is unable to make any of the available appointment slots, prompt them to call the Repairs Contact Centre, quoting the reference *(aka work order reference number)*.

cannot-make-appointment


dgmstuart (approved):
LGTM 🚢

PR #69, opened 2017-12-01
(236) Upgrade Rails and Nokogiri

dgmstuart:
Upgrading rails also upgraded Nokogiri


leeky (approved):
Works for me! :shipit:

PR #72, opened 2017-12-04
(269) Add Rack::Timeout

leeky:


dgmstuart:
As discussed: I've modified this to coerce the environment variable into an integer.

dgmstuart (approved):
LGTM 🚢

PR #73, opened 2017-12-04
(241) Dont show invalid appointments

leeky:
After requesting available appointment slots from the API, filter the results as follows:

 - Only return 'best' slots
 - Don't return any appointments for the current day or in the past
 - Filter all day appointments (those longer than 6 hours)
 - Return a maximum of 15 appointment slots (controllable via the `APPOINTMENT_LIMIT` environment variable)

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -34,11 +34,10 @@ class AppointmentsController < ApplicationController
   end
 
   def available_appointments
-    appointments = HackneyApi.new.list_available_appointments(
-      work_order_reference: work_order_reference
-    )
-
-    appointments.map { |a| AppointmentPresenter.new(a) }
+    AppointmentFetcher
+      .new
+      .call(work_order_reference: work_order_reference)
+      .map { |appointment| AppointmentPresenter.new(appointment) }
   end
 
   def book_appointment_save_into_answer_store
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -36,7 +36,10 @@ class AppointmentsController < ApplicationController
   def available_appointments
     AppointmentFetcher
       .new
-      .call(work_order_reference: work_order_reference)
+      .call(
+        work_order_reference: work_order_reference,
+        limit: ENV.fetch('APPOINTMENT_LIMIT', 15).to_i
+      )
       .map { |appointment| AppointmentPresenter.new(appointment) }
   end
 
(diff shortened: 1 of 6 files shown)


app/services/appointment_fetcher.rb
```diff
@@ -0,0 +1,16 @@
...
+    @work_order_reference = work_order_reference
+
+    appointments = available_appointments
+    appointments.map { |appointment| AppointmentPresenter.new(appointment) }
+  end
+
+  private
+
+  def available_appointments
+    @available_appointments ||= HackneyApi.new.list_available_appointments(
```
dgmstuart: Since `available_appointments` is only called once inside this class, what do we get from memoizing it?

app/services/appointment_fetcher.rb
```diff
@@ -0,0 +1,16 @@
+class AppointmentFetcher
+  def call(work_order_reference:)
+    @work_order_reference = work_order_reference
+
+    appointments = available_appointments
```
dgmstuart: What does this variable assignment give us?

app/controllers/appointments_controller.rb
```diff
@@ -34,7 +34,10 @@ def work_order_reference
   end
 
   def available_appointments
-    AppointmentFetcher.new.call(work_order_reference: work_order_reference)
+    AppointmentFetcher
+      .new
+      .call(work_order_reference: work_order_reference)
+      .map { |appointment| AppointmentPresenter.new(appointment) }
```
dgmstuart: The previous commit mapped these inside `AppointmentFetcher` and this one puts it back in the controller. Is it worth fixing that up?

app/services/appointment_fetcher.rb
```diff
@@ -3,7 +3,8 @@ def call(work_order_reference:)
     @work_order_reference = work_order_reference
 
     appointments = available_appointments
-    appointments.map { |appointment| AppointmentPresenter.new(appointment) }
+    appointments = filter_before_tomorrow(appointments)
+    filter_long_slots(appointments)
```
dgmstuart: Would this read better as something like:

appointments.map {}
  .select { |a| after_tomorrow(a) }
  .reject { |a| long_slot(a) }

or even

appointments.map {}
  .select(&:after_tomorrow)
  .reject(&:long_slot)  

Maybe not worth it

dgmstuart (approved):
Generally good: a few potential improvements, but happy to go as is.

app/controllers/appointments_controller.rb
leeky: :+1:

leeky:
I've tidied up a couple of places, but left the filter chaining part alone, as it was taking longer than I wanted to spend to refactor it.

PR #74, opened 2017-12-04
(267) Log requests we make to the API

leeky:
Enable Faraday's built-in logger, so that it's easier to debug issues with the API.


dgmstuart (approved):
LGTM 👍

PR #70, opened 2017-12-04
(263) Make site usable in Internet Explorer 8

leeky:
Conditionally load an IE8-friendly stylesheet, HTML5 shiv and `background-size` polyfill. This makes IE8 behave itself.

### Before

before-ie8

### After

after-ie8

PR #75, opened 2017-12-04
(237) Nice error pages for application errors

dgmstuart:
Approach:

- Use the Rails default approach of rendering static pages:
  - Benefit: real errors get raised and logged to rollbar
  - Downside: some duplication of markup and styles between error pages
    and normal app pages
- Use the same css as the application error page which is on AWS
- Make the static pages minimal (no information which might get out of
  date) and link to a separate page, served by the app, which lists the
  contact phone numbers.
- Include back links on the 404 and 422 pages, but not on the 500 page:
  not sure it makes sense there?

**NOTE** The css needs updating to work on ie8 as per https://github.com/LBHackney-IT/maintain-my-home/pull/70#pullrequestreview-80963966


dgmstuart:
Rebasing

dgmstuart:
Reviewed for copy by @lexijax

spec/features/residents_see_nice_errors_spec.rb
```diff
@@ -1,6 +1,39 @@
 require 'rails_helper'
 
 RSpec.feature 'Error pages' do
+  scenario 'when an application error occurs' do
+    runtime_error = RuntimeError.new("They're coming outta the walls! They're coming outta the goddamn walls!")
```
tahb, line 5: 😆

spec/features/residents_see_nice_errors_spec.rb
```diff
@@ -1,6 +1,39 @@
...
+    allow(StartForm).to receive(:new)
+      .and_raise runtime_error
+
+    allow(Rails.logger).to receive(:error)
+    allow(Rollbar).to receive(:error)
+
+    visit '/'
+    expect { click_on 'Start' }.to raise_error runtime_error # ...and rely on rails default routing to render the error
```
tahb, line 13: This test looks to assert when visiting the root path the StartForm is called and raises. Would it be more useful to assert that that we see content from our new 500 error page?

spec/features/residents_see_nice_errors_spec.rb
```diff
@@ -29,7 +29,7 @@
 
   scenario 'when a page was not found' do
     visit '/422'
-    expect(page).to have_content "We're sorry, but the change you wanted was rejected"
+    expect(page).to have_content "We're really sorry"
```
tahb: While this checks we are looking at the right page it seems to me that the meaningful message on the page is in fact:

> Something seems to have gone wrong while reporting your repair.

It would be more helpful in this case if our tests told us if this part was accidentally removed, over the header.

public/404.html
```diff
@@ -1,67 +1,51 @@
...
-    <div>
-      <h1>The page you were looking for doesn't exist.</h1>
-      <p>You may have mistyped the address or the page may have moved.</p>
+  <main>
+    <div id="content" role="main">
+      <h1>
+        Sorry - the page you were looking for doesn't seem to exist! 
+      </h1>
+      <p>
+        You can try going
```
tahb, line 40: On the 422 page this identical paragraph is made bold.

tahb (approved):
Tricky to review the contents of minified CSS but the pages all seem styled nicely.

Couple of minor suggestions for you that you can choose to leave undone.

spec/features/residents_see_nice_errors_spec.rb
dgmstuart: We can't: Capybara shows errors in test mode.

Agreed this test case isn't great. What I'm trying to make sure is that generic errors don't get somehow silently rescued, and instead bubble all the way to the top. Maybe this test doesn't actually give us that?

I think it's possible to temporarily disable error rendering, but I think that ended up seeming like a massive faff

PR #76, opened 2017-12-05
(278) Use legacy YouTube embed code

leeky:
The video sharing code you get from youtube.com only supports new
browsers, as they no longer support Flash video embeds.

However, using the old embed code still works in modern browsers, but
additionally provides a fallback for IE8 and older - which shows the
video thumbnail with a button that links to the YouTube video on their
site directly:

youtube-ie8


dgmstuart (approved):
LGTM 🚢

PR #78, opened 2017-12-05
(fix) Request YouTube videos over SSL

leeky:
Whilst this worked in local development, once it was deployed browsers
blocked the YouTube content, as it was being requested over HTTP rather
than HTTPS.


dgmstuart (approved):
LGTM - seen it working over Leeky's shoulder

PR #79, opened 2017-12-20
(273) Include last question in the description

dgmstuart:
Identify these as being the ones with a 'desc' key (i.e. it's
explicitly the last question), or if it has no 'next' or 'page' (i.e.
it's implicitly the last question.

The fact that this is implicit feels fragile, but I guess it makes it
easier to write the questions.yml

Made last question mandatory because if it's not there then something
has gone really wrong. Unfortunately this results in a bunch of tests
failing, but I think that's more of a sign of our tests not being
specific enough.

### DIFF
--- a/app/models/repair_params.rb
+++ b/app/models/repair_params.rb
@@ -6,6 +6,7 @@ class RepairParams
   def problem_description
     lines = [description]
     lines << "Room: #{room}" if room.present?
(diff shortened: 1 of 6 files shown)


dgmstuart (commented):
Nobody available to review, but I've looked it over again and I'm happy

isobelcroot (approved):

PR #80, opened 2017-12-21
(275) Leaseholders cannot report repairs

dgmstuart:
The API 'get property' endpoint will shortly return a new `maintainable`
boolean.

When this is `false`, it is likely the property belongs to a leaseholder or
is a non-Hackney Council managed address.

In this case, redirect the user to a page telling them to call the Repairs
Contact Centre for further assistance.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -9,5 +9,4 @@
 # Offense count: 4
 Metrics/AbcSize:
   Exclude:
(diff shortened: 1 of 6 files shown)


dgmstuart (commented):
Nobody available to review this, but I've looked it over again and I'm happy.

isobelcroot (approved):

PR #81, opened 2017-12-21
(284) remove communal/outside repairs

dgmstuart:
These can't be raised against properties (need to be against blocks or
estates instead), so we can't handle them yet. Instead bump the 'which
room?' question to be the first, and change the wording accordingly.

Hopefully reverting this commit will allow us to reinstate them, but I'm
not 100% sure it'll be that simple.


dgmstuart (commented):
Nobody available to review this. I've looked it over again and am happy with it.

isobelcroot (approved):

PR #86, opened 2018-01-09
(261) Validate all static pages compile correctly

leeky:
Visit each static page in turn, and ensure they respond with an `HTTP 200`. Previously, an error crept into one of the page templates, and it wasn't noticed until a user flagged it.


leeky:
[Check that all pages are valid](https://trello.com/c/JUIJMYK5/261-check-that-all-pages-are-valid)

evilstreak (approved):
👍

PR #88, opened 2018-02-08
(306) Users can be sent to One Account when our service is disabled

leeky:
In order for us to test 'Report a Repair' in live, add a new route `/landing' that` can be linked to from the button on `https://hackney.gov.uk/report-a-repair`.

The normal behaviour is to immediately redirect the user to our root path, so they can begin using the service.

In the event that we have to disable our service for some reason (via the `service_disabled` feature flag we already have in place) `/landing` will instead redirect users to the existing One Account service.

Finally, in case One Account's URL is changed for some reason, I've added an optional `ONE_ACCOUNT_URL` environment variable. Setting this allows users to be sent to this URL instead.

This puts the control of the redirect back to One Account in our hands, so that it can be performed without requiring any changes to the main hackney.gov.uk web site.


evilstreak (approved):
👍 Looks good

PR #89, opened 2018-02-09
(134) Add BETA banner to the top of all pages

leeky:
This follows the pattern used by GOV.UK, and includes a link to a static page which explains to users what a 'beta' service is.

I've taken the copy directly from the GOV.UK guidance, with some minor renames.

---
beta-banner


evilstreak (approved):
👍

PR #90, opened 2018-02-12
(309) Hide the 'beta' banner when the site is disabled

leeky:
Disabling the site replaces all pages with a warning that the service is unavailable. Unfortunately, the beta banner's link would be broken in this case.

I've therefore removed the beta banner from all pages when the service is disabled.


evilstreak (approved):
👍 (though the weird negative messes with my head)

leeky:
[Hide 'beta' banner when service disabled with feature flag](https://trello.com/c/VWfa6acg/309-hide-beta-banner-when-service-disabled-with-feature-flag)

PR #91, opened 2018-02-12
(308) Allow landing page bouncing to be controlled separately

leeky:
Previously, the behaviour of the redirect on '/launch' was linked to the `service_disabled` feature flag. This would prevent testing the site in production after the link on hackney.gov.uk goes live.

Now, you can redirect users back to the One Account via `/launch`, whilst allowing the service to be tested independently.

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -2,7 +2,7 @@ class LandingPageController < ApplicationController
   skip_before_action :check_service_status
 
   def index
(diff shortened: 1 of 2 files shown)


evilstreak (approved):
👍

leeky:
[Allow site to be disabled without affecting '/launch' redirect](https://trello.com/c/hJEUFeK8/308-allow-site-to-be-disabled-without-affecting-launch-redirect)

PR #92, opened 2018-02-12
(310) Show non-best slots if no best slots are available

leeky:
Whilst testing, we were offered *no* appointment slots because no 'best' slots are available. This change shows non-best slots in this case instead.


evilstreak (commented):
👍 Looks good

evilstreak (approved):

PR #93, opened 2018-02-15
(298) BUG - key not found 'last_question'

leeky:
I've fixed this bug by always saving the current question and answer to the `last_question` key in the session, instead of using logic to try and only write to this key when we reach a leaf node in the decision tree, which didn't always work properly.

### DIFF
--- a/app/services/question_saver.rb
+++ b/app/services/question_saver.rb
@@ -40,19 +40,11 @@ class QuestionSaver
 
   def persist_last_question(form)
     answer = @question.answer_data(form.answer)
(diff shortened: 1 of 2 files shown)


leeky:
['key not found: "last_question"' error after entering contact details.](https://trello.com/c/EmnTzGcn/298-key-not-found-lastquestion-error-after-entering-contact-details)

evilstreak (approved):
👍 Nice; deleting unnecessary logic is the best kind of PR!

PR #95, opened 2018-02-21
(316) Don't make excessive API calls

leeky:
In a couple of places, we needed either the `repair_request_reference` or `work_order_reference` for the newly created repair. Memoise the call to `HackneyApi#get_repair` so that we re-use the response rather than requesting it multiple times.

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -26,11 +26,7 @@ class AppointmentsController < ApplicationController
   end
 
   def work_order_reference
-    Repair.new(
-      HackneyApi.new.get_repair(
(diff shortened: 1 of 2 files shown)


leeky:
[Don't make excessive API calls](https://trello.com/c/SnuDCTfj/316-dont-make-excessive-api-calls)

evilstreak (approved):
👍 Memoising these API calls is definitely the right thing to do, and these changes are good to go.

I have one concern which is probably beyond the scope of this PR: I don't think a Presenter object should be making API calls – I think it violates expectations about what a presenter does. What do you think about passing a `Repair` object into the `Confirmation` initialiser rather than having it call the API itself?

PR #96, opened 2018-02-23
New description format

evilstreak:
Based on user testing with DLO operatives we're updating the description to make it more useful for them. The new description format is:

```
Room: Bathroom
Problem with: Bath

Freeform resident description goes here…
```

The commit messages have more details about each change.

### DIFF
--- a/app/models/repair_params.rb
+++ b/app/models/repair_params.rb
@@ -4,11 +4,10 @@ class RepairParams
   end
 
   def problem_description
-    lines = [description]
-    lines << "Room: #{room}" if room.present?
-    lines << "Last question: #{last_question}"
-    lines << "Callback requested: between #{callback_time}" if callback_time
-    lines.compact.join("\n\n")
+    [
+      metadata,
+      description,
+    ].compact.join("\n\n")
   end
 
   def property_reference
@@ -37,6 +36,15 @@ class RepairParams
 
   private
 
+  def metadata
+    data = []
+    data << "Room: #{room}" if room_specified?
+    data << "Problem with: #{problem}" if problem.present?
+    data << "Callback requested: between #{callback_time}" if callback_time
+
+    data.compact.join("\n").presence
+  end
+
   def description
     @answers.fetch('describe_repair').fetch('description').tap do |desc|
       return 'No description given' if desc.blank?
@@ -57,5 +65,13 @@ class RepairParams
     return if time.blank?
 
     Callback::TimeSlot.new(time)
+  end
+
+  def room_specified?
+    room.present? && room != 'Other'
+  end
+
+  def problem
+    @answers.dig('problem', 'problem')
   end
 end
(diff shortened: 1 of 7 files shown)


spec/services/question_saver_spec.rb
```diff
@@ -39,6 +39,28 @@
     end
 
     context 'when there is a room on the question' do
+      it 'persists form data to the selected answer store' do
```
leeky: Is this test and the one below the wrong way round? (i.e. this one is in the `room` context, but tests that the `problem` is persisted?)

app/services/question_saver.rb
```diff
@@ -19,6 +19,7 @@ def persist_answers(form)
     persist_room(form)
     persist_sor_code(form)
     persist_last_question(form)
+    persist_problem(form)
```
leeky, line 22: Is it worth removing the `persist_last_question` method, if we're no longer using this within the metadata part of the description..?

app/models/repair_params.rb
```diff
@@ -4,11 +4,15 @@ def initialize(answers)
...
-    lines << "Last question: #{last_question}"
-    lines << "Callback requested: between #{callback_time}" if callback_time
-    lines.compact.join("\n\n")
+    metadata = []
+    metadata << "Room: #{room}" if room_specified?
+    metadata << "Problem with: #{problem}" if problem.present?
+    metadata << "Callback requested: between #{callback_time}" if callback_time
+
+    [
+      metadata.compact.join("\n").presence,
```
leeky: `presence` is new to me :+1:

spec/services/create_repair_spec.rb
```diff
@@ -85,43 +88,6 @@
     end
   end
 
-  it 'posts a default description, if none was provided' do
```
leeky, line 88: Any reason why this was removed?

leeky (changes requested):
A few minor comments, but otherwise looks like a nice solution :)

leeky:
@evilstreak Could you add a card for this on Trello and attach the PR to it?

spec/services/question_saver_spec.rb
evilstreak: Ack, yes, I copied the block and then changed the wrong one. Thanks for catching that

spec/services/create_repair_spec.rb
evilstreak: Because it's no longer relevant compared to when it was first added – the relevant commit message has more detail.

app/services/question_saver.rb
evilstreak: It's probably worth getting rid of, yes. I'm happy to do that as a follow up.

evilstreak:
Updated for the test mis-labelling and the rubocop failures, and I've added a Trello card with a link to the PR.

leeky (approved):
LGTM! :shipit:

PR #97, opened 2018-02-23
Add message for no available appointments

evilstreak:
The API will try to find some appointments available in the next 4 weeks - if it can't, it will return an empty list. For that case we want to show the resident a useful page, rather than asking them to pick an appointment from a list of none.

Hopefully this page will never appear in LIVE, as there should always be some available appointments within the month, but if it does it needs to be useful for the resident using it.

The page looks like this:

screen shot 2018-02-23 at 15 30 10

### DIFF
--- a/app/views/appointments/none_available.html.haml
+++ b/app/views/appointments/none_available.html.haml
@@ -1,0 +1,24 @@
+.grid-row
+  %section#confirmation.column-two-thirds
+    %h1
+      Booking your appointment
+
+    %p
+      Sorry, we can't book an appointment for you online right now.
+
+    .box-highlight.lede
+      Your reference number is
+      = succeed "." do
+        %strong= @work_order_reference
+
+    %p
+      %strong
+        Please call us on
+        = repairs_contact_centre_telephone_number
+        to arrange an appointment
+      and quote your reference number
+      = succeed "." do
+        %strong= @work_order_reference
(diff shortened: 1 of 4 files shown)


app/views/appointments/none_available.html.haml
```diff
@@ -0,0 +1,22 @@
...
+
+    %p
+      %strong
+        Please call us on
+        = repairs_contact_centre_telephone_number
+        to arrange an appointment
+      and quote your reference number
+      = succeed "." do
+        %strong= @work_order_reference
+      We're open Monday to Friday, 8am to midnight and Saturday, 9am to 1pm.
```
leeky: We added the opening hours to the i18n file, because it changed a few times during the project, i.e:

```haml
%p
  We're open
  = succeed '.' do
    = t('rcc_opening_hours')
```

Probably worth doing this here as well.

leeky (approved):
One comment about opening hours, otherwise LGTM :shipit:

app/views/appointments/none_available.html.haml
evilstreak: Oh, interesting. I copied these from the `_contact_phone_numbers` partial, so I should probably update that too!

evilstreak: Done.

PR #98, opened 2018-02-23
Always require freeform problem description

evilstreak:
Previously a problem that we had an SOR code for had an optional freeform description. To try to get all the information a DLO operative needs to do job we're changing it to require the description whether we have an SOR code or not.

See also: #96

### DIFF
--- a/app/models/describe_repair_form.rb
+++ b/app/models/describe_repair_form.rb
@@ -3,5 +3,5 @@ class DescribeRepairForm
 
   attr_accessor :description
 
-  validates :description, length: { maximum: 500 }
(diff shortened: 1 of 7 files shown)


leeky (approved):
LGTM! :shipit:

Although it's no longer used, it might be worth keeping in the optional field code for SimpleForm. It was a bit of a faff to get working.

evilstreak:
To summarise an IRL discussion: I agree on the optional field thing. The reason I deleted it was because of the 98% code coverage restriction. I'm going to drop it to 95% and then get rid of the commit that deletes the optional field.

PR #99, opened 2018-02-23
Update failing rspec feature

evilstreak:
When the no-appointments screen was added it was based off a commit that didn't include the problem description updates. The spec worked fine on its own branch, but once it got merged into `develop` it failed because the description generated by the app had changed.

This updates the expectations inline with the new description format.


leeky (approved):
LGTM! :shipit:

PR #100, opened 2018-02-26
Remove callback "within one working day" promise

evilstreak:
Remove the timeframe promised for the callback after discussion with the RCC. We can revisit the promised timeframe in the future once we understand the impact of the service.


leeky (approved):
LGTM! :shipit:

PR #101, opened 2018-03-23
(318) Upgrade vulnerable gems

leeky:
The following gems were recently flagged as having security vulnerabilities, so I've upgraded them:

 - loofah
 - nokogiri
 - rack-protection
 - rails-html-sanitizer


LBHSPreston (approved):
Looks fine to me.

PR #104, opened 2018-05-10
Change Hackney API URLs

james:
As part of some reorganisation of HackneyAPI, the root path of the API is now prefixed with /hackneyrepairs/

I have been unable to do any tests on this locally, as I don't have access to the network the API server is restricted to yet.


LBHSPreston (approved):
Change location of the Repairs api module.

PR #105, opened 2018-05-10
Retry changing API root URL

james:
As is expected when I am unable to test anything, the last commit didn't work. This is my second, and hopefully final attempt.


LBHSPreston (approved):
Change api endpoint urls.

PR #106, opened 2018-05-22
Report another repair

james:
Add new link at end of process to restart process, but retain address and contact
details.
If address is present, then new address confirmation page
New method for removing address information if user wants to change
If contact details present, then prefill form with them

Resolves trello card [40-add-another-repair-feature](https://trello.com/c/jSWJPrFR)

New "Add another repair" link on confirmation page:
1b17740d-4e12-4241-8375-666c11414813

New address confirmation page:
13e31cc1-6594-4e58-8577-c139a85b5028

### DIFF
--- a/app/views/address_searches/confirm_address.html.haml
+++ b/app/views/address_searches/confirm_address.html.haml
@@ -1,0 +1,18 @@
+= back_link
+%h1
+  Confirm address
+
+%p
+  Is this the address for repair?
+
+%p
+  %aside#progress
+    = @address['address']
+    = @address['postcode']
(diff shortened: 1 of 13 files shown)


app/views/address_searches/confirm_address.html.haml
```diff
@@ -0,0 +1,18 @@
+= back_link
+%h1
+  Confirm address
+
+%p
+  If this the address the repair needs to be made at?
```
deberny: Could we change it to 'Is this the address for repair?'

deberny (approved):

PR #118, opened 2018-06-04
Improved journey for damp, leak, mould

deberny:
Resolves trello card
https://trello.com/c/ulPBgzNt/75-ko-journey-for-damp-vs-leak

### DIFF
--- a/db/questions.yml
+++ b/db/questions.yml
@@ -20,7 +20,7 @@ kitchen_problem:
       problem: Cupboards or Worktop
 
     - text: Damp or Mould
-      page: damp_and_mould
+      next: damp_or_mould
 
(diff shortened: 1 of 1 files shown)


deberny:
@james I've kept the old damp leak page in case we need to use it.

This is the new journey. There will be need to add meaningful free text description depending on the problem a user ends up with. Amy is working on it.

deberny:
[KO journey for Damp VS Leak](https://trello.com/c/ulPBgzNt/75-ko-journey-for-damp-vs-leak)

james (approved):
I'm not sure if you want to wait for the description bits before merging, so I'll leave that to you, but LGTME!

PR #121, opened 2018-06-14
New SOR for leaking pipe, removed SOR from condensation

deberny:
Resolves Trello card [/79-damp-leak-mould-add-the-right-sor-codes](https://trello.com/c/ONXKUlKy)

Confirmed SOR codes for Damp and Moud.

A leaking pipe has a new SOR code, that also will need to be included in Hackney API.

Condensation does not book a surveyor, instead we give the user a helpful information on what causes condensation and give them a link to tips on how to prevent condensation available on Hackney Council website.

Condensation problem page:
screen shot 2018-06-14 at 11 40 29

### DIFF
--- a/app/views/pages/condensation_problem.html.haml
+++ b/app/views/pages/condensation_problem.html.haml
@@ -1,0 +1,24 @@
+= back_link
+
+%h1
+  What to do if you have a problem with condensation
+
+%p
+  Condensation is the moisture build up caused by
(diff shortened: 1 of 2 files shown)


deberny:
[Damp/Leak/Mould add the right SOR codes](https://trello.com/c/ONXKUlKy/79-damp-leak-mould-add-the-right-sor-codes)

james (approved):

PR #125, opened 2019-02-19
Add docker-compose

mallorydxw:
Run the project with: `docker-compose up`

Rebuild the image (after dependencies change) with: `docker-compose
build`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM ruby:2.4
+RUN apt-get update -qq && apt-get install -y nodejs postgresql-client
+RUN mkdir /myapp
+WORKDIR /myapp
+COPY Gemfile /myapp/Gemfile
+COPY Gemfile.lock /myapp/Gemfile.lock
+RUN bundle install
+COPY . /myapp
+
+# Add a script to be executed every time the container starts.
+COPY entrypoint.sh /usr/bin/
+RUN chmod +x /usr/bin/entrypoint.sh
+ENTRYPOINT ["entrypoint.sh"]
+EXPOSE 3000
+
+# Start the main process.
+CMD ["rails", "server", "-b", "0.0.0.0"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,20 @@
+version: '3'
+services:
+  db:
+    image: postgres:11.2
+    volumes:
+      - ./tmp/db:/var/lib/postgresql/data
+  redis:
+    image: redis:5.0
+  web:
+    build: .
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ruby:2.4
+RUN apt-get update -qq && apt-get install -y nodejs postgresql-client
```
tahb, line 2: Does this application need `nodejs` or `postgresql-client` to run?

If it doesn't, I guess these have been added as a convenience for development only. As this is still a Dockerfile that may end up being one day used in production, it seems to me that the safest option would be to install this package only when needed, which I presume would be very rarely:

```
docker-compose run --rm web bash
$ apt-get update -qq && apt-get install -y nodejs postgresql-client
$ … use tools to investigate
```

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY . /myapp
+
+# Add a script to be executed every time the container starts.
+COPY entrypoint.sh /usr/bin/
+RUN chmod +x /usr/bin/entrypoint.sh
+ENTRYPOINT ["entrypoint.sh"]
+EXPOSE 3000
+
+# Start the main process.
+CMD ["rails", "server", "-b", "0.0.0.0"]
```
tahb, line 17: I wonder if this binding is needed, does Docker network magic make this available at localhost anyway?

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '3'
+services:
+  db:
+    image: postgres:11.2
```
tahb, line 4: On checking the postgres running on live at the moment is only `9.4`, having this set higher locally could make things harder to debug.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    volumes:
+      - ./tmp/db:/var/lib/postgresql/data
+  redis:
+    image: redis:5.0
```
tahb, line 8: On checking the postgres running on live at the moment is only `4.0.14`, having this set higher locally could make things harder to debug.

tahb:
Thank you for contributing this change which I presume helped you when you came to work on this repository.

I understand the benefits of using Docker Compose to keeping our host machines clean of all the dependencies and to have it all removed once we're done. That said there is a cost in having  both docker and non-docker routes options around to manage the app locally, it can cause confusion where a new joiner isn't sure which to use and either risk going out of date depending on the last developers preference.

Personally I would be in favour of removing all non-docker startup instructions and moving to Docker Compose in full however given where this project is in its lifecycle, it's unlikely to require much more development so we may no longer reap the benefits of doing this.

I think it is better to leave this project with a single reliable set of start up documentation and tooling unless the situation changes.

PR #137, opened 2019-11-27
Call Google Drive calls in the background

pezholio:
We've been getting a few issues coming out of Rollbar telling us that requests to create a repair are timing out because of Google drive. After a bit of back and forth, we decided that the best way to mitigate against this was to use workers to process these jobs. I decided that the best option was Delayed Job. I considered Sidekiq, as we already have Redis setup, but it's not recommended to mix data storage and job processing in the same Redis setup. We've got a Postgres Database hanging around not doing anything, so it seemed like the next obvious chose was DJ.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ git_source(:github) do |repo_name|
 end
 
 gem 'business_time'
+gem 'delayed_job_active_record'
 gem 'faker' # Use in all environments until we have a real API
 gem 'faraday'
 gem 'faraday_middleware'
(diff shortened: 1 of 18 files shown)


Alex-Farley:
@pezholio Do you know why the ci/circleci check is showing as expected?

pezholio:
@Alex-Farley Looks like something to do with the Circle CI config is bust. Nothing has run for a couple of weeks now. I can't dig into it any more because I don't have admin access to the repo

mec:
I don't think anything is wrong with CircleCI as a dependabot PR opened later was fine. 

This work is on a fork which is why the CI did not run, I've updated the settings for this project in CircleCI to run on forks as well, although a better solution is to work on branches on this repo – this will un-block this PR from being merged.

I am not sure why an earlier dependabot PR failed, but that one is in CircleCI and re-running the workflow made it go green.

PR #140, opened 2020-01-21
(fix) Do not reset answers when there are none

mec:
We saw Rollbar errors from visits to `questions/start` with the
`keep_address` param set to `true` when there was not session with
answers to reset.

https://rollbar.com/dxw/Hackney-Repairs/items/77/occurrences/90813887448/


mec:
At over a year old, it feels like this will never get a review so I am closing it.

